PR #46, opened 2026-03-16
Add configurable /tmp tmpfs size with ByteSize type

JAORMX:
## Summary
- Wire propolis `WithTmpSize` through all brood-box layers so users can configure guest `/tmp` tmpfs size
- Introduce `ByteSize` type in config domain that accepts human-readable sizes (`"512m"`, `"2g"`) and bare integers (MiB) in YAML
- Guest init reads `/etc/propolis-vm.json` written by the host to apply the configured size at boot

## Config example
```yaml
defaults:
  tmp_size: "2g"       # 2 GiB for all agents
agents:
  claude-code:
    tmp_size: "4g"     # 4 GiB for this agent
```

Bare integers remain backward compatible (`tmp_size: 512` = 512 MiB).

## Changes
- **`pkg/domain/config/bytesize.go`**: `ByteSize` type with `ParseByteSize`, `encoding.TextUnmarshaler`/`TextMarshaler` (no yaml.v3 import in domain layer)
- **`pkg/domain/config/config.go`**: `DefaultsConfig.TmpSize` and `AgentOverride.TmpSize` → `ByteSize`; split clamping; `Merge()` converts to `uint32` at agent boundary
- **`pkg/domain/agent/agent.go`**: `DefaultTmpSize uint32` field
- **`pkg/domain/vm/vm.go`**: `TmpSizeMiB uint32` in `VMConfig`
- **`pkg/sandbox/sandbox.go`**: Wires `TmpSizeMiB` through `RunOpts` → override → `VMConfig`
- **`internal/infra/vm/runner.go`**: Passes `TmpSizeMiB` to `propolis.WithTmpSize`
- **`cmd/bbox-init/main.go`**: Reads vmconfig, passes to `boot.WithTmpSize`

## Dependencies
- Depends on stacklok/propolis#44 — `go.mod` has a temporary `replace` directive until a propolis release is cut

## Test plan
- [x] `task test` passes
- [ ] Remove `replace` directive after propolis release and verify


pkg/domain/config/bytesize.go
```diff
@@ -0,0 +1,111 @@
...
+	n, err := strconv.ParseUint(numStr, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("invalid byte size %q: %w", s, err)
+	}
+
+	var mib uint64
+	switch suffix {
+	case "", "m", "mi", "mib":
+		mib = n
+	case "g", "gi", "gib":
```
jhrozek, line 100: nit: `n * 1024` can silently overflow `uint64` for very large inputs. For example, `"18014398509481985g"` parses `n = 18014398509481985`, and `n * 1024` wraps around to `1024`. The post-check against `math.MaxUint32` then passes, silently returning `ByteSize(1024)` instead of an error.

Practical impact is low (wrapped values are always ≤ MaxTmpSize), but it's a correctness bug — the user specified ~16 EiB and got 1 GiB.

```suggestion
	case "g", "gi", "gib":
		if n > math.MaxUint64/1024 {
			return 0, fmt.Errorf("invalid byte size %q: exceeds maximum (%d MiB)", s, uint64(math.MaxUint32))
		}
		mib = n * 1024
```

pkg/domain/config/config.go
```diff
@@ -412,11 +432,15 @@ func MergeConfigs(global, local *Config) *Config {
 	if local.Defaults.Memory > 0 {
 		result.Defaults.Memory = local.Defaults.Memory
 	}
```
jhrozek, line 434: Question: Should workspace-local `.broodbox.yaml` be able to *increase* the tmpfs size beyond the global default? CPUs and Memory use the same free-override pattern, but I wonder if there should be an upper limit from workspace config (e.g., capped at the global/agent default) to prevent a malicious repo from inflating resource allocation via its `.broodbox.yaml`. The `MaxTmpSize` cap at 64 GiB may be sufficient, but wanted to flag the question.

cmd/bbox-init/main.go
```diff
@@ -60,7 +61,15 @@ func main() {
 	stopReaper := reaper.Start(logger)
 	defer stopReaper()
 
-	shutdown, err := boot.Run(logger, boot.WithSSHAgentForwarding(true))
+	vmCfg, vmCfgErr := vmconfig.Read()
```
jhrozek, line 64: Minor: `vmCfg` is used on line 71 even when `vmconfig.Read()` returns an error. This works today because Go returns a zero-value `Config{}`, but if propolis ever changes the return type to a pointer (`*Config`), this becomes a nil deref inside PID 1 with no recovery. Consider making the fallback explicit:

```suggestion
	vmCfg, vmCfgErr := vmconfig.Read()
	if vmCfgErr != nil {
		logger.Warn("failed to read vm config, using defaults", "error", vmCfgErr)
		vmCfg = vmconfig.Config{}
	}
```

pkg/sandbox/sandbox.go
```diff
@@ -85,6 +85,9 @@ type RunOpts struct {
 	// LogLevel sets the hypervisor log verbosity (0=off, 5=trace).
 	LogLevel uint32
 
+	// TmpSizeMiB overrides the /tmp tmpfs size in MiB (0 = use agent/global default).
```
jhrozek, line 88: Question: Is YAML-only configuration intentional for `TmpSizeMiB`, or is a `--tmp-size` CLI flag planned as a follow-up? CPUs and Memory have CLI flags, so the asymmetry stood out. If YAML-only is by design, a brief comment here would help.

jhrozek (commented):
Nice, clean feature addition — follows the existing CPUs/Memory patterns well and respects DDD boundaries. A few questions and a small correctness nit below.

jhrozek (approved):

PR #57, opened 2026-03-17
Add `bbox config init` command

JAORMX:
## Summary

- Add `bbox config init` subcommand that generates `~/.config/broodbox/config.yaml` with every configuration option documented as YAML comments
- All values are commented out so the file is safe to ship as-is and parses to a zero-value config
- Respects `--config` for a custom path and `--force` to overwrite an existing file
- Update CLAUDE.md with guidance to keep the template in sync when config fields change

## Test plan

- [x] Unit tests cover: file creation, parent directory creation, existing-file guard, `--force` overwrite, template parses to zero-value `Config`, all config sections present
- [ ] Manual: `task build && ./bin/bbox config init` writes the file, inspect contents
- [ ] Manual: re-run without `--force` → errors; re-run with `--force` → overwrites

### DIFF
--- a/internal/infra/config/writer.go
+++ b/internal/infra/config/writer.go
@@ -22,7 +22,8 @@ var defaultConfigTemplate = `# Brood Box configuration
 #   # Number of vCPUs (0 = agent default).
 #   cpus: 0
 #
-#   # RAM in MiB (0 = agent default).
+#   # RAM for the VM. Accepts human-readable values like "4g" or "512m".
+#   # Bare integers are treated as MiB. Zero uses the agent default.
 #   memory: 0
 #
 #   # Size of /tmp tmpfs inside the VM. Accepts human-readable values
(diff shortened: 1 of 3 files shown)


internal/infra/config/writer.go
```diff
@@ -0,0 +1,187 @@
...
+			return fmt.Errorf("checking config file: %w", err)
+		}
+	}
+
+	dir := filepath.Dir(path)
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return fmt.Errorf("creating config directory: %w", err)
+	}
+
+	if err := os.WriteFile(path, []byte(defaultConfigTemplate), 0o600); err != nil {
```
jhrozek, line 184: hmm, the permission will only be 0600 if the file is created, the call doesn't chown if the file exists (--force was used). Do we want to add an explicit chown?

internal/infra/config/writer.go
```diff
@@ -0,0 +1,187 @@
...
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+// defaultConfigTemplate is a fully commented YAML template documenting every
+// Config field. All values are commented out so the file is safe to ship as-is.
+var defaultConfigTemplate = `# Brood Box configuration
+# All values below are commented out and show their defaults.
```
jhrozek, line 17: should authz and config be added here?

jhrozek (dismissed):
two nits inline non-blocking plus https://github.com/stacklok/brood-box/issues/62 filed for a follow-up

jhrozek (approved):

PR #60, opened 2026-03-17
Use ByteSize for memory config

JAORMX:
## Summary

- Change `DefaultsConfig.Memory` and `AgentOverride.Memory` from `uint32` to `ByteSize`, matching the existing `TmpSize` pattern
- Update `--memory` CLI flag from `Uint32Var` to `StringVar` with `ParseByteSize()`, so users write `--memory 4g` instead of `--memory 4096`
- Update `MaxMemory` constant to `ByteSize` and adapt `clampResources()` and `Merge()` accordingly
- Update docs and config examples to use human-readable format (`"4g"`, `"512m"`)

Bare integers in YAML config are still accepted (backward compatible via `ByteSize.UnmarshalText`).

## Test plan

- [x] `task fmt && task lint` — 0 issues
- [x] `task test` — all tests pass
- [ ] Manual: verify `bbox claude-code --memory 4g` works
- [ ] Manual: verify `memory: "4g"` in config.yaml works
- [ ] Manual: verify bare integer `memory: 4096` still works (backward compat)

### DIFF
--- a/cmd/bbox/main.go
+++ b/cmd/bbox/main.go
@@ -65,7 +65,7 @@ func main() {
 func rootCmd() *cobra.Command {
 	var (
 		cpus              uint32
-		memory            uint32
+		memory            string
 		wsPath            string
 		sshPort           uint16
 		cfgPath           string
@@ -107,7 +107,7 @@ Supported agents: claude-code, codex, opencode
 
 Example:
   bbox claude-code
-  bbox codex --cpus 4 --memory 4096
+  bbox codex --cpus 4 --memory 4g
   bbox opencode --workspace /path/to/project
   bbox claude-code --review
   bbox claude-code --review --exclude "*.log" --exclude "tmp/"
@@ -159,7 +159,7 @@ Example:
 	}
 
 	cmd.Flags().Uint32Var(&cpus, "cpus", 0, "Number of vCPUs (0 = agent default)")
-	cmd.Flags().Uint32Var(&memory, "memory", 0, "RAM in MiB (0 = agent default)")
+	cmd.Flags().StringVar(&memory, "memory", "", "RAM for the VM, e.g. 4g or 512m (0 = agent default)")
 	cmd.Flags().StringVar(&tmpSize, "tmp-size", "", "Size of /tmp tmpfs inside the VM, e.g. 512m or 2g (0 = agent default)")
 	cmd.Flags().StringVar(&wsPath, "workspace", "", "Workspace directory to mount (default: current directory)")
 	cmd.Flags().Uint16Var(&sshPort, "ssh-port", 0, "Host SSH port (0 = auto-pick)")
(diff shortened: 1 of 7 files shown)


pkg/domain/config/config.go
```diff
@@ -565,9 +568,11 @@ func Merge(a agent.Agent, override AgentOverride, defaults DefaultsConfig) agent
 	}
 
 	// Clamp resource values to configured maximums.
-	result.DefaultCPUs, result.DefaultMemory = clampResources(
-		result.DefaultCPUs, result.DefaultMemory,
+	clampedCPUs, clampedMem := clampResources(
+		result.DefaultCPUs, ByteSize(result.DefaultMemory),
```
jhrozek, line 572: This `ByteSize(result.DefaultMemory)` cast feels a bit off — we're converting `uint32 → ByteSize` just to call `clampResources`, then immediately `.MiB()` it back to `uint32`. It works because both represent MiB under the hood, but it's the kind of thing that quietly breaks if `ByteSize` ever changes its internal representation.

Should we just make `agent.Agent.DefaultMemory` (and `DefaultTmpSize`) a `ByteSize` too? That'd kill all the `.MiB()` / `ByteSize()` ping-pong in `Merge()` and keep the domain types consistent. The `uint32` conversion would only happen at the VM boundary where the hypervisor actually needs raw MiB.

Not blocking — just wondering if it's worth a quick follow-up.

cmd/bbox/main.go
```diff
@@ -159,7 +159,7 @@ Example:
 	}
 
 	cmd.Flags().Uint32Var(&cpus, "cpus", 0, "Number of vCPUs (0 = agent default)")
-	cmd.Flags().Uint32Var(&memory, "memory", 0, "RAM in MiB (0 = agent default)")
+	cmd.Flags().StringVar(&memory, "memory", "", "RAM for the VM, e.g. 4g or 512m (0 = agent default)")
```
jhrozek, line 162: Nit: this is a string flag now (default `""`), so `(0 = agent default)` is a bit misleading. `--tmp-size` on the next line doesn't have this parenthetical either.

```suggestion
	cmd.Flags().StringVar(&memory, "memory", "", "RAM for the VM, e.g. 4g or 512m (empty = agent default)")
```

jhrozek (commented):
Clean change, consistent with the `tmp_size` pattern. A couple of small things inline.

jhrozek (approved):
Looks good — the inline comments are just nits, fine to fix later.

pkg/domain/config/config.go
JAORMX: I'll address this in a follow-up

cmd/bbox/main.go
JAORMX: I'll address this in a follow-up

PR #63, opened 2026-03-18
Narrow per-agent MCP override type and add authz support

### LINKED ISSUE #62: Clarify and align per-agent MCP override semantics (mcp.config vs mcp.authz)
While reviewing PR #57 (not merged at the time of writing), I noticed a pre-existing inconsistency in per-agent MCP override handling.

`agents.<name>.mcp` is currently typed as a full `MCPConfig`, so the config schema/loader accepts both:

- `agents.<name>.mcp.config`
- `agents.<name>.mcp.authz`

However, runtime behavior does not appear to treat those fields consistently:

- per-agent `mcp.config` is applied when resolving the effective MCP config
- per-agent `mcp.authz` appears to be parsed but ignored

Relevant code paths:

- `pkg/domain/config/config.go`
  - `AgentOverride.MCP` is `*MCPConfig`
- `pkg/sandbox/sandbox.go`
  - `resolveMCPConfig` copies `enabled`, `group`, `port`, and `config`, but not `authz`

Why this matters:

- The public config shape suggests per-agent `mcp.authz` is supported.
- In practice, it does not seem to affect runtime behavior.
- This creates schema/runtime drift and makes the config surface harder to reason about.
- The new `config init` PR makes this more visible because it introduces generated config documentation, but the underlying behavior looks older than that PR.

Expected outcome:

We should make schema, runtime behavior, and generated documentation agree.

Possible fixes:

1. Support per-agent `mcp.authz` fully
   - merge it in `resolveMCPConfig`
   - add tests covering the effective config behavior

2. Do not support per-agent `mcp.authz`
   - narrow `AgentOverride.MCP` to a dedicated override type
   - expose only intentionally supported fields
   - make docs/template reflect that narrower shape

3. In either case
   - ensure generated config docs match actual supported per-agent MCP fields
   - add tests for per-agent MCP override resolution

Notes:

Workspace-local `.broodbox.yaml` security stripping is a separate concern; this issue is about the global config schema/runtime mismatch for per-agent MCP overrides.

JAORMX:
## Summary

- Replace `AgentOverride.MCP` (`*MCPConfig`) with a new `MCPAgentOverride` type exposing only `Enabled` and `Authz` — makes unsupported fields (`Group`, `Port`, `Config`) unrepresentable at compile time
- Export `MergeMCPAuthzConfig` and add tighten-only per-agent authz merging in the composition root before `VMCPProvider` construction
- Simplify `resolveMCPConfig` to Enabled-only overrides, removing dead Group/Port/Config code that was parsed but never had runtime effect
- Update `warnLocalConfigOverrides` to warn about per-agent `Authz` instead of the removed fields

## Test plan

- [x] `task fmt && task lint` — 0 issues
- [x] `task test` — all tests pass with race detector
- [ ] CI passes (test, lint, build matrix)
- [ ] Manual: config with `agents.claude-code.mcp.authz.profile: observe` parses correctly and the tighter profile propagates

## Backward compatibility

Go's `yaml.v3` silently drops unknown fields. Users with existing per-agent `mcp.group`/`mcp.port`/`mcp.config` in YAML will see those fields silently ignored — same as today since they never had runtime effect. The Go API change (`AgentOverride.MCP` type) is breaking for external consumers, acceptable at v0.0.x.

Fixes #62

### DIFF
--- a/pkg/domain/config/config.go
+++ b/pkg/domain/config/config.go
@@ -365,6 +365,20 @@ func StricterMCPAuthzProfile(a, b string) string {
 	return a
 }
 
+// MCPAgentOverride holds the subset of MCP settings that can be
+// overridden per-agent. Only Enabled (gate) and Authz (tighten-only)
+// have runtime effect at the per-agent level.
+type MCPAgentOverride struct {
+	// Enabled controls whether the MCP proxy is active for this agent.
+	// nil means "no override" (inherit global setting), unlike
+	// MCPConfig.Enabled where nil defaults to true.
+	Enabled *bool `yaml:"enabled,omitempty"`
+
+	// Authz overrides the MCP authorization profile for this agent.
+	// Tighten-only: can only make the profile stricter, not more permissive.
+	Authz *MCPAuthzConfig `yaml:"authz,omitempty"`
+}
+
 // IsEnabled returns whether the MCP proxy is enabled.
 // Defaults to true when Enabled is nil.
 func (m MCPConfig) IsEnabled() bool {
@@ -457,7 +471,8 @@ type AgentOverride struct {
 	AllowHosts []EgressHostConfig `yaml:"allow_hosts,omitempty"`
 
 	// MCP overrides the global MCP proxy settings for this agent.
-	MCP *MCPConfig `yaml:"mcp,omitempty"`
+	// Only Enabled (gate) and Authz (tighten-only) are supported.
+	MCP *MCPAgentOverride `yaml:"mcp,omitempty"`
 
 	// SettingsImport overrides the global settings import for this agent.
 	SettingsImport *SettingsImportConfig `yaml:"settings_import,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/domain/config/config.go
```diff
@@ -510,10 +524,15 @@ func MergeConfigs(global, local *Config) *Config {
...
+				if ga, ok := result.Agents[k]; ok && ga.MCP != nil {
+					globalAgentAuthz = ga.MCP.Authz
+				}
 				sanitized := *v.MCP
-				sanitized.Config = nil
-				sanitized.Authz = nil
+				sanitized.Authz = MergeMCPAuthzConfig(globalAgentAuthz, v.MCP.Authz)
 				v.MCP = &sanitized
 			}
 			result.Agents[k] = v
```
jhrozek, line 611: **Bug: Per-agent security fields silently dropped by wholesale replacement**

`result.Agents[k] = v` replaces the entire global `AgentOverride` with the local one. This means **any** local agent override — even just `cpus: 4` — stomps the global per-agent MCP authz and egress profile.

**Scenario:**
```yaml
# Global config (trusted)
agents:
  claude-code:
    mcp:
      authz:
        profile: observe

# Workspace .broodbox.yaml (untrusted)
agents:
  claude-code:
    cpus: 4
```

Result: `claude-code` has `cpus: 4` but `MCP.Authz` is nil — the `observe` restriction is gone.

The same issue affects per-agent `EgressProfile` (wholesale-replaced, not tighten-only like top-level `Defaults.EgressProfile`).

**Suggestion:** Extract a `mergeAgentOverride(global, local AgentOverride) AgentOverride` function that merges field-by-field:
- Resource/identity fields (`Image`, `CPUs`, `Memory`, etc.): local overrides when non-zero
- `EgressProfile`: tighten-only via `egress.Stricter` (matching top-level pattern)
- `AllowHosts`: additive (matching `Network.AllowHosts` pattern)
- `MCP.Authz`: tighten-only via `MergeMCPAuthzConfig`
- `MCP.Enabled`: local overrides global

This follows the existing patterns (`mergeGitConfig`, `mergeRuntimeConfig`, `MergeMCPAuthzConfig`) and ensures adding new security fields to `AgentOverride` in the future has a clear merge checklist.

jhrozek (changes requested):
## Review Summary

**1 HIGH, 1 LOW** finding (confidence ≥ 8 only).

The type narrowing from `*MCPConfig` to `*MCPAgentOverride` is a solid security improvement — making illegal states unrepresentable. DDD layer boundaries are respected throughout. `MergeMCPAuthzConfig` is well-designed and thoroughly tested.

However, the agent merge loop in `MergeConfigs` does wholesale replacement of `AgentOverride` values, which silently drops security-relevant fields set by the operator in global config.

### LOW: Dead data flow in composition root

`cmd/bbox/main.go:655-656` writes `sandboxCfg.MCP.Config` and `sandboxCfg.MCP.Authz` but the sandbox layer never reads them (`pkg/sandbox/` has zero references to `.MCP.Authz` or `.MCP.Config`). Authz is resolved entirely at the composition root before `VMCPProvider` construction. These writes are harmless but potentially misleading for future readers. Consider removing them or adding a comment explaining they exist for SDK consumers.

jhrozek (approved):

PR #94, opened 2026-03-27
Fix git worktree object database in VM snapshots

### LINKED ISSUE #49: Worktree snapshots lack object database — git shows "No commits yet"
## Problem
When running an agent in a git worktree workspace, `git status` inside the VM works without errors but shows "No commits yet" with all files untracked. This is because worktree object databases live in the main repo's `.git/objects/`, which is outside the workspace and not mounted in the VM.

## Current state
As of #48, worktree snapshots have:
- Sanitized `.git/config` (readable, credentials stripped)
- `HEAD` (symbolic ref from the worktree's gitdir)
- Empty `objects/` and `refs/` directories

This is enough for `git status` to run without errors, but git can't resolve any refs.

## Options considered
1. **Mount main repo's `.git/objects/` read-only** into the VM — gives full history, objects are immutable so read-only mount is safe. Requires wiring the resolved commondir path through sanitizer → VM runner.
2. **Copy pack files** — works but expensive for large repos.
3. **Synthetic initial commit** — `git init && git add . && git commit` programmatically. Agent can track its own changes but loses original history.
4. **`objects/info/alternates`** — requires a valid in-VM path to the main repo's objects, so depends on option 1.

jhrozek:
## Summary

When a workspace is a git worktree, the object database lives in the main repository's `.git/objects/` — outside the workspace and not included in the snapshot. This causes git inside the VM to show "No commits yet" with all files appearing untracked.

Fixes #49

## Approach

Mount the main repo's `.git/objects/` as a **read-only** virtiofs share and reconstruct a proper `.git/` directory in the snapshot that references it via `objects/info/alternates`.

### Domain layer
- Extend `SnapshotPostProcessor` to return `PostProcessResult` with optional mount requests and diff-exclude patterns
- Add `ExtraMounts` field to `VMConfig`

### Infrastructure
- Add `WorktreeProcessor` that detects worktrees, reconstructs `.git/` in the snapshot with `objects/info/alternates` pointing to the guest mount, and returns mount + exclude requests
- Add `InjectMountConfig` rootfs hook writing `/etc/broodbox-mounts.json`
- Wire extra virtiofs mounts in VM runner

### Guest
- Add `mountExtras` in bbox-init reading the mount config and mounting additional virtiofs shares between boot and seccomp hardening

### Application
- Collect post-processor results in sandbox orchestrator, wire mounts into VMConfig, compose diff matcher with extra excludes

## Security hardening

- **Path traversal prevention**: Validates gitdir path contains `/.git/worktrees/` before proceeding; rejects symlinks via `os.Lstat` in file copy
- **Defense-in-depth**: Verifies resolved git directory contains HEAD before treating it as a git repo
- **Config sanitization**: Credentials stripped from `.git/config` via existing `SanitizeConfig()`
- **Host path not exposed**: Guest mount config JSON excludes HostPath — guest only sees tag, mount point, and read-only flag

## Trade-off

Agent file changes persist through normal flush+review but git commits do not persist (the reconstructed `.git` is excluded from diff). This avoids the complexity of mapping reconstructed git state back to the original worktree structure.

**Known limitation**: Host-side virtiofs mount has no read-only enforcement (go-microvm's `VirtioFSMount` lacks a `ReadOnly` field). Read-only is enforced guest-side via `MS_RDONLY` + seccomp blocking `mount`/`umount` syscalls.

## Test plan

- [x] `TestWorktreeProcessor_ExternalWorktree` — full reconstruction verified
- [x] `TestWorktreeProcessor_DetachedHEAD` — raw SHA HEAD preserved
- [x] `TestWorktreeProcessor_CraftedGitdir` — path traversal rejected
- [x] `TestWorktreeProcessor_SymlinkInGitdir` — symlink not followed
- [x] `TestWorktreeProcessor_ConfigIsSanitized` — credentials stripped
- [x] `TestWorktreeProcessor_NotAWorktree` / `NoGitEntry` / `MalformedGitFile` / `MissingObjects` — graceful no-ops
- [x] `TestComposeMatcher` — prefix matching verified, no false positives on `.github`/`.gitignore`
- [x] `task lint` — 0 issues
- [ ] Manual: `bbox claude-code` in a git worktree → `git log` shows history

### DIFF
--- a/cmd/bbox-init/main.go
+++ b/cmd/bbox-init/main.go
@@ -86,6 +86,10 @@ func main() {
 			"error", err)
 	}
 
+	if err := mountExtras(logger); err != nil {
+		logger.Warn("failed to mount extras", "error", err)
+	}
+
 	if err := harden.ApplySeccomp(); err != nil {
 		logger.Error("seccomp filter failed", "error", err)
 		halt()
--- a/cmd/bbox-init/mounts.go
+++ b/cmd/bbox-init/mounts.go
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux
+
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"syscall"
+	"time"
+)
+
+// guestMountEntry matches the JSON written by InjectMountConfig on the host.
+type guestMountEntry struct {
+	Tag       string `json:"tag"`
+	GuestPath string `json:"guest_path"`
+	ReadOnly  bool   `json:"read_only"`
+}
+
+// mountConfigPath is the guest path where the host writes extra mount config.
+const mountConfigPath = "/etc/broodbox-mounts.json"
+
+// mountExtras reads /etc/broodbox-mounts.json and mounts each virtiofs share.
+// If the config file does not exist, it returns nil (no extra mounts needed).
+func mountExtras(logger *slog.Logger) error {
+	data, err := os.ReadFile(mountConfigPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return fmt.Errorf("reading mount config: %w", err)
+	}
+
+	var entries []guestMountEntry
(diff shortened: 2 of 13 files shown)


jhrozek:
Draft,because I haven't had time to test this properly but this is what I used to run bbox in worktrees

jhrozek:
no longe a draft

JAORMX (approved):

PR #124, opened 2026-04-16
Inject GOMEMLIMIT into guest VM to prevent Go tool OOM kills

JAORMX:
## Summary

- Go tools spawned by agents (e.g. `golangci-lint`) don't auto-detect VM memory limits — the guest runs its own kernel with no cgroup boundary, so the Go GC has no soft ceiling and grows unbounded until the OOM killer fires
- Adds `GoMemLimitPercent` to the agent definition (same pattern as the existing `NodeHeapPercent` / `NODE_OPTIONS` cap) and injects `GOMEMLIMIT` into `/etc/sandbox-env`
- Sets 70% for all built-in agents — with a 10 GB VM this yields `GOMEMLIMIT=7168MiB`, leaving headroom for Node.js and system overhead

## Context

Observed `golangci-lint` (3.6 GB RSS) getting OOM-killed inside a 10 GB VM, which left the SSH session frozen. The kernel's `oom_score_adj` fired because no Go GC soft limit was in place.

## Test plan

- [x] `task fmt` — clean
- [x] `task lint` — 0 issues
- [x] `task test` — all passing


jhrozek (approved):
it would be cool to be able to set env vars like this through sub-agents. But great, this is probably why I've been seeing occasionaly OOMs as well.

PR #137, opened 2026-04-22
Add --workspace-mode=direct to skip snapshot isolation

JAORMX:
## Summary

- Adds `--workspace-mode=<snapshot|direct>` (default `snapshot`) and a matching `workspace.mode` global config key. Direct mode mounts the workspace read-write inside the VM with no COW snapshot, no diff, no review, and no git config sanitization.
- Per-workspace `.broodbox.yaml` cannot enable direct mode: the merge is tighten-only (mirrors the egress and MCP authz patterns), so a repo can only force snapshot, never widen to direct. A warning is printed when a workspace config attempts it.
- First-run `--yes` acknowledgement is persisted at `\$XDG_STATE_HOME/broodbox/direct-mode-ack`; subsequent direct-mode runs do not require the flag. A startup stderr banner prints every direct-mode run.
- `--workspace-mode=direct` with `--review` or `--exclude` on the same invocation is a hard error (silent-ignore would be a data-loss footgun). `review.enabled: true` from global config is warned-and-ignored when direct wins.
- `--review` stays exactly as it is today: still an opt-in, still meaningful in snapshot mode, no rename and no deprecation.

## Safety notes

Direct mode loses: git config sanitization, worktree post-processor, hash-verified flush, sensitive-path classification, snapshot post-processors. It keeps: egress policy, credential seeding, settings injection, MCP authz. The README has a new "Workspace modes" section calling this out.

Architecture: `SandboxRunner` already has nil-guards on `WorkspaceCloner` / `Reviewer` / `Flusher` / `Differ` at every call site, so direct mode is wired by leaving those deps nil and setting `Snapshot.Enabled=false` on `RunOpts`. No new implementations were needed.

## Test plan

- [x] `task fmt` clean
- [x] `task lint` clean
- [x] `task test` passes (added 5 merge cases + `StricterWorkspaceMode`, `IsValidWorkspaceMode`, `Config.Validate`, and `ResolvedWorkspaceMode` tests in `pkg/domain/config/`)
- [x] CLI validation smoke-tested: `--workspace-mode=bogus`, `--workspace-mode=direct --review`, `--workspace-mode=direct --exclude=foo`, `--workspace-mode=direct` without `--yes` all error clearly
- [ ] Manual end-to-end run with `bbox claude-code --workspace-mode=direct --yes` on a throwaway workspace
- [ ] Manual run with a `.broodbox.yaml` containing `workspace.mode: direct` to confirm the warning fires and the mode is ignored

### DIFF
--- a/pkg/domain/config/config.go
+++ b/pkg/domain/config/config.go
@@ -22,6 +22,10 @@ type Config struct {
 	// Defaults specifies default resource limits.
 	Defaults DefaultsConfig `yaml:"defaults"`
 
+	// Workspace configures how the workspace is exposed to the VM
+	// (snapshot isolation vs direct pass-through).
+	Workspace WorkspaceConfig `yaml:"workspace"`
+
 	// Review configures workspace snapshot isolation.
 	Review ReviewConfig `yaml:"review"`
 
@@ -62,6 +66,10 @@ type Config struct {
 // has its own Validate). Add more here as new classes of footgun
 // emerge.
 func (c *Config) Validate() error {
+	if !IsValidWorkspaceMode(c.Workspace.Mode) {
+		return fmt.Errorf("workspace.mode %q: valid values are %v",
+			c.Workspace.Mode, ValidWorkspaceModes())
+	}
 	for name, override := range c.Agents {
 		if err := agent.ValidateEnvForwardPatterns(override.EnvForward); err != nil {
 			return fmt.Errorf("agents.%s.%w", name, err)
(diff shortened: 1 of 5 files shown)


pkg/domain/config/config.go
```diff
@@ -466,6 +474,83 @@ type ReviewConfig struct {
...
+		return false
+	}
+}
+
+// ResolvedWorkspaceMode returns the effective mode, mapping "" to the default.
+func (w WorkspaceConfig) ResolvedWorkspaceMode() string {
+	if w.Mode == "" {
+		return WorkspaceModeSnapshot
+	}
+	return w.Mode
```
jhrozek, line 525: `ResolvedWorkspaceMode()` is only called from its own unit test — `run()` in `cmd/bbox/main.go:572-581` inlines the same `"" → snapshot` default manually. Two sources of truth for the normalization.

Suggest either dropping the method, or (preferred) calling `cfg.Workspace.ResolvedWorkspaceMode()` from `run()` so the domain method is the single source of truth. Non-blocking.

jhrozek (dismissed):
LGTM. One non-blocking inline nit about `ResolvedWorkspaceMode()` being dead code outside its own test — either remove it or wire it in as the single source of truth for the default.

jhrozek (approved):

PR #139, opened 2026-04-27
Fix handling of read-only source file cloning

Kensan:
When invoking `bbox` in a working directory with read-only file(s), the operation will fail with a "context canceled" error message. These changes fix handling of such files by opening the destination file with write permissions and changing them to match the source file permissions after a successful copy operation.

Additionally, swap the error evaluation order in `CreateSnapshot` in order to show underlying I/O error messages.

### DIFF
--- a/internal/infra/workspace/clone.go
+++ b/internal/infra/workspace/clone.go
@@ -37,7 +37,8 @@ func copyFile(src, dst string) error {
 	// Strip setuid/setgid/sticky — only preserve rwx permissions.
 	mode := info.Mode().Perm()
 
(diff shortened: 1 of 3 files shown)


JAORMX (approved):
Thanks for the fix! Confirmed the bug reproduces (FICLONE creates dst at the source's `0o400` mode, fails on tmpfs, and the `copyFile` fallback can't reopen the read-only dst — `EACCES` gets masked as "context canceled" via errgroup propagation). The reorder makes the underlying error visible, and opening at `0o600` + post-`Chmod` is the right fix. macOS path is unaffected (`unix.Clonefile` preserves perms in-kernel). LGTM.

PR #151, opened 2026-05-13
Fix: prevent recursive snapshot self-copy when cache dir falls within workspace

allen-munsch:
## Problem

When the workspace is the home directory (`~/`) or any ancestor of `~/.cache/broodbox/snapshots/`, the snapshot temp dir is created **inside** the workspace tree being walked.

`filepath.WalkDir` calls the callback for a directory **before** reading its entries. When the callback fires for the temp dir, `os.MkdirAll(destPath, ...)` creates a nested copy of the snapshot dir inside itself. `WalkDir` then calls `readDir` on the source directory, discovers the freshly-planted nested directory, and recurses — creating an infinite loop.

Each iteration adds another segment to the path until the OS's `ENAMETOOLONG` limit is hit:

```
Error: creating workspace snapshot: cloning files: creating destination:
open .../.cache/broodbox/snapshots/.sandbox-snapshot-123/
     broodbox/snapshots/.sandbox-snapshot-123/
     broodbox/snapshots/.sandbox-snapshot-123/
     ...(repeating)...
     /libkrunfw.so.5: file name too long
```

## Fix

Skip the entire `snapshotBaseDir` tree in the `WalkDir` callback. When the path equals or is under the snapshot cache directory, return `filepath.SkipDir`. This prevents both the current tmpDir and any stale snapshots from being walked.

## Test

Added `TestCreateSnapshot_SelfRecursionPrevention` that places the snapshot cache inside the workspace (simulating the CWD=~/ scenario), pre-populates cache-like files, and verifies the snapshot completes without recursion and that the cache directory is not copied into the snapshot.

edit: signed off by me, overviewed, pr written and opened via AI, looked legit (also a real issue i ran into)

### DIFF
--- a/internal/infra/workspace/snapshot.go
+++ b/internal/infra/workspace/snapshot.go
@@ -111,6 +111,15 @@ func (c *FSWorkspaceCloner) CreateSnapshot(ctx context.Context, workspacePath st
 			return nil
 		}
 
+		// Skip the snapshot cache directory to prevent recursive self-copy
+		// when the snapshot base dir falls within the workspace tree.
+		if path == c.snapshotBaseDir || strings.HasPrefix(path, c.snapshotBaseDir+string(os.PathSeparator)) {
+			if d.IsDir() {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+
 		destPath := filepath.Join(tmpDir, relPath)
 
 		// Handle symlinks first (metadata-only, fast — stays synchronous).
--- a/internal/infra/workspace/snapshot.go
+++ b/internal/infra/workspace/snapshot.go
@@ -53,7 +53,7 @@ type FSWorkspaceCloner struct {
 func NewFSWorkspaceCloner(cloner FileCloner, snapshotBaseDir string, logger *slog.Logger) *FSWorkspaceCloner {
 	return &FSWorkspaceCloner{
 		cloner:          cloner,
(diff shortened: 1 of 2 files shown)


jhrozek (dismissed):
Clean, focused fix — the separator-aware prefix check is correct and the test directly exercises the bug scenario. Approving.

**Nit / non-blocking suggestion** — `NewFSWorkspaceCloner` stores `snapshotBaseDir` as-is. The guard added by this PR relies on exact string equality (`path == c.snapshotBaseDir`), where `path` always comes out of `filepath.WalkDir` in normalized form. All current callers go through `filepath.Join` so they're fine, but an SDK consumer passing a trailing-slash path would silently defeat the guard. A one-liner fix at construction time makes it robust:

```go
// internal/infra/workspace/snapshot.go — NewFSWorkspaceCloner
snapshotBaseDir: filepath.Clean(snapshotBaseDir),
```

`filepath.Clean` is idempotent on already-clean paths so there's no cost in the common case.

allen-munsch:
thanks added the nit fix

i see its mutating into an assignment, but i guess the alternatives would add per directory overhead, so added it at the constrctor

jhrozek (approved):

jhrozek (approved):

jhrozek:
@allen-munsch thanks, merged and tagged a new release 🚀

PR #198, opened 2026-06-26
Clean up orphaned VM data dirs from crashed VMs

### LINKED ISSUE #87: Clean up orphaned `rootfs-work` dirs from crashed VMs
## Problem

When a VM crashes or the parent `bbox` process is killed (SIGKILL, OOM, etc.), `WithCleanDataDir()` never runs. The COW-cloned rootfs at `~/.config/broodbox/vms/<name>/data/rootfs-work/` survives — potentially ~800MB per orphaned VM.

Brood-box already has stale cleanup for two similar cases:
- `infravm.CleanupStaleLogs()` removes old VM log directories using PID sentinel files
- `infraws.CleanupStaleSnapshots()` removes old workspace snapshots

But orphaned `rootfs-work/` dirs inside VM data directories are not covered.

## Proposal

Extend the existing stale cleanup to also handle orphaned VM data directories (which contain the `rootfs-work/` clone).

### Approach

go-microvm already persists VM state in `<dataDir>/state.json` with the runner PID and an `active` flag. The cleanup logic should:

1. On startup, scan `~/.config/broodbox/vms/*/data/state.json`
2. For each entry where `active: true`, check if the PID is still alive
3. If the PID is dead, the VM was orphaned — remove the entire data directory (including `rootfs-work/`)

go-microvm's `terminateStaleRunner()` already does steps 1-2 for killing orphaned processes. The data dir cleanup is the missing step 3 — `cleanDataDir` handles it for the *current* VM's data dir, but not for data dirs from *other* crashed VMs.

### Where this lives

This could be:
- A new `CleanupStaleVMData()` function in `internal/infra/vm/` alongside the existing cleanup helpers
- Called from the composition root (`cmd/bbox/main.go`) on startup, next to the existing `CleanupStaleLogs` and `CleanupStaleSnapshots` calls

## Notes

- The PID-liveness check is already battle-tested in the existing stale cleanup code
- This is safe for concurrent VMs: each VM has its own named data directory, and we only clean dirs whose PID is confirmed dead
- This pairs with the image cache GC work — cleaning both the rootfs cache and orphaned rootfs clones covers both sides of the disk waste

JAORMX:
## Summary

Fixes #87 — orphaned COW rootfs clones from crashed VMs leak disk indefinitely.

When a VM crashes or `bbox` is killed (SIGKILL, OOM) before `WithCleanDataDir()` runs, the COW-cloned rootfs at `~/.config/broodbox/vms/<name>/data/rootfs-work/` survives — potentially hundreds of MB per orphan, accumulating unbounded across crashed runs.

`CleanupStaleLogs` already removes a whole VM directory when its bbox **sentinel** PID is dead, but no sentinel is written for runs that pass `--log-file`, so those data clones leak.

## Approach

Add `CleanupStaleVMData`, keyed off go-microvm's per-VM **state file** (`state.Manager.Load()`, which reads `go-microvm-state.json` without locking) rather than the bbox sentinel:

- Scan `~/.config/broodbox/vms/*/data/`.
- A data dir whose state is `active: true` but whose recorded **runner PID is dead** was orphaned → remove it.
- A **live** runner PID is left untouched, so it's safe to run at startup alongside concurrent VMs (each VM has its own uniquely named dir). A recycled PID now hosting an unrelated live process is conservatively treated as alive and skipped; the orphan is reclaimed on a later run.
- After removing the data dir, best-effort removes the now-empty parent (reclaiming sentinel-less `--log-file` runs) via a non-recursive `Remove`, which only succeeds on an empty dir — so it never deletes lingering logs or a sentinel that `CleanupStaleLogs` still owns.

Wired into the composition root next to `CleanupStaleLogs` / `CleanupStaleSnapshots`.

> Note: the issue referenced `state.json`; the actual go-microvm state file is `go-microvm-state.json`. Reusing `state.Manager.Load()` means we don't hardcode the filename or schema.

## Tests

New table of cases in `cleanup_test.go` (state fixtures written via the canonical `state` package):
- orphaned (active + dead PID) → data dir **and** empty parent removed
- live runner (active + our own PID) → preserved
- inactive state (clean shutdown) → preserved
- orphan **with** parent logs → only data dir removed, parent left for log cleanup
- no state file → preserved
- nonexistent vms dir → no panic

Verified: `task lint` (0 issues), full `task test` passes (incl. `-race`).

### DIFF
--- a/cmd/bbox/main.go
+++ b/cmd/bbox/main.go
@@ -519,9 +519,12 @@ func run(parentCtx context.Context, agentName string, flags runFlags) error {
 	// Clean up stale snapshot dirs from previous crashes.
 	infraws.CleanupStaleSnapshots(snapDir, logger)
 
-	// Clean up stale VM log directories from previous crashes.
+	// Clean up stale VM log directories and orphaned data dirs (COW rootfs
+	// clones) left behind by crashed runs.
 	if home, homeErr := os.UserHomeDir(); homeErr == nil {
-		infravm.CleanupStaleLogs(filepath.Join(home, ".config", "broodbox", "vms"), logger)
+		vmsDir := filepath.Join(home, ".config", "broodbox", "vms")
+		infravm.CleanupStaleLogs(vmsDir, logger)
+		infravm.CleanupStaleVMData(vmsDir, logger)
 	}
 
 	// Build registry: start with built-in clients, then layer in any
--- a/cmd/bbox/main.go
+++ b/cmd/bbox/main.go
@@ -521,10 +521,28 @@ func run(parentCtx context.Context, agentName string, flags runFlags) error {
 
 	// Clean up stale VM log directories and orphaned data dirs (COW rootfs
 	// clones) left behind by crashed runs.
-	if home, homeErr := os.UserHomeDir(); homeErr == nil {
+	//
+	// BBOX_KEEP_VM_DATA=1 disables startup stale-dir reclamation in addition
+	// to gating Stop(): a crashed run under that env var leaves Active==true
+	// + a dead PID — exactly what the sweep removes — so without this gate
+	// the documented debug procedure (set the env var → re-run → inspect)
+	// would silently nuke the dir the user wanted to inspect.
+	if os.Getenv("BBOX_KEEP_VM_DATA") == "1" {
+		logger.Debug("BBOX_KEEP_VM_DATA set, skipping stale VM directory cleanup")
+	} else if home, homeErr := os.UserHomeDir(); homeErr == nil {
 		vmsDir := filepath.Join(home, ".config", "broodbox", "vms")
-		infravm.CleanupStaleLogs(vmsDir, logger)
-		infravm.CleanupStaleVMData(vmsDir, logger)
+		// Background the sweep: data/rootfs-work/ can be GBs of small files,
+		// and an unbounded synchronous RemoveAll would serialize before VM
+		// boot on every interactive invocation. Each VM dir is uniquely
+		// named and the current run's vmName is passed as skipName, so there
+		// is no TOCTOU against the current run.
+		go func() {
+			start := time.Now()
+			logger.Debug("starting stale VM directory cleanup", "vms_dir", vmsDir)
+			infravm.CleanupStaleVMDirs(vmsDir, vmName, logger)
+			logger.Debug("finished stale VM directory cleanup",
+				"vms_dir", vmsDir, "elapsed", time.Since(start))
+		}()
 	}
 
 	// Build registry: start with built-in clients, then layer in any
--- a/internal/infra/vm/cleanup.go
+++ b/internal/infra/vm/cleanup.go
@@ -12,6 +12,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/stacklok/go-microvm/state"
+
 	"github.com/stacklok/brood-box/internal/infra/process"
 )
 
@@ -115,6 +117,87 @@ func CleanupStaleLogs(vmsDir string, logger *slog.Logger) {
 			"path", dirPath, "pid", pid, "reason", reason)
 		if err := os.RemoveAll(dirPath); err != nil {
 			logger.Error("failed to remove stale VM log directory", "path", dirPath, "error", err)
+		}
+	}
+}
+
+// vmDataSubdir is the per-VM data subdirectory (under ~/.config/broodbox/vms/<name>/)
+// that holds the go-microvm state file and the COW rootfs clone (rootfs-work/).
+const vmDataSubdir = "data"
+
+// CleanupStaleVMData removes orphaned VM data directories left behind when a
+// VM crashed or its bbox process was killed (SIGKILL, OOM, ...) before
+// WithCleanDataDir() could run. The COW rootfs clone under rootfs-work/ can be
+// hundreds of MB per orphan, so left unattended these accumulate unbounded.
+//
+// Each VM's go-microvm state file records the runner PID and an "active" flag.
+// A data dir whose state is still active but whose runner PID is dead was
+// orphaned and can be safely reclaimed. This complements CleanupStaleLogs:
+// that helper keys off the bbox sentinel and removes the whole VM directory,
+// but no sentinel is written when a run uses --log-file, so the data clone
+// would otherwise leak. Keying off the runner state covers that case too.
+//
+// A live runner PID is left untouched, so this is safe to run at startup
+// alongside concurrent VMs (each VM has its own uniquely named directory). A
+// recycled PID that now hosts an unrelated live process is conservatively
+// treated as alive and skipped; the orphan is reclaimed on a later run once
+// the PID is free.
+func CleanupStaleVMData(vmsDir string, logger *slog.Logger) {
+	entries, err := os.ReadDir(vmsDir)
+	if err != nil {
+		// Directory may not exist yet on first run — not an error.
+		if os.IsNotExist(err) {
+			return
+		}
+		logger.Warn("failed to scan for stale VM data directories", "error", err)
+		return
+	}
+
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+
+		vmDir := filepath.Join(vmsDir, entry.Name())
+		dataDir := filepath.Join(vmDir, vmDataSubdir)
+
+		// Load reads go-microvm's state file without locking. A missing
+		// file (no VM ever ran here, or it was already cleaned) yields a
+		// default inactive state, which we skip below.
+		st, loadErr := state.NewManager(dataDir).Load()
+		if loadErr != nil {
+			logger.Debug("skipping VM dir with unreadable state",
+				"path", dataDir, "error", loadErr)
+			continue
+		}
+
+		// Not an orphan: no active VM recorded, or no runner PID to check.
+		if !st.Active || st.PID <= 0 {
+			continue
+		}
+
+		if process.IsAlive(st.PID) {
+			logger.Debug("skipping VM data dir owned by running runner",
+				"path", dataDir, "pid", st.PID)
+			continue
+		}
+
+		logger.Warn("removing orphaned VM data directory",
+			"path", dataDir, "pid", st.PID, "reason", "runner-dead")
+		if err := os.RemoveAll(dataDir); err != nil {
+			logger.Error("failed to remove orphaned VM data directory",
+				"path", dataDir, "error", err)
+			continue
+		}
+
+		// Drop the parent VM dir too if it is now empty — this reclaims
+		// directories from --log-file runs that left no sentinel or logs
+		// here. A non-recursive Remove only succeeds on an empty dir, so it
+		// never deletes lingering logs or a sentinel that CleanupStaleLogs
+		// still owns.
+		if err := os.Remove(vmDir); err != nil && !os.IsNotExist(err) {
+			logger.Debug("VM dir not empty after data cleanup; leaving for log cleanup",
+				"path", vmDir)
 		}
 	}
 }
--- a/internal/infra/vm/cleanup.go
+++ b/internal/infra/vm/cleanup.go
@@ -4,6 +4,7 @@
 package vm
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"log/slog"
@@ -11,6 +12,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/stacklok/go-microvm/state"
 
@@ -27,6 +29,24 @@ const LogSentinel = ".bbox-sentinel"
 // forcing bbox to load a multi-GiB file into memory at startup by
 // planting a giant sentinel.
 const maxSentinelSize = 4096
+
+// maxStateSize is a generous upper bound on any single file inside a
+// VM data directory. A well-formed go-microvm state file is a few
+// hundred bytes; 4 KiB caps a planted giant (CWE-400 hardening) and
+// mirrors the sentinel discipline, applied to LoadAndLock's unbounded
+// os.ReadFile underpinnings.
+const maxStateSize = 4096
+
+// runnerBinaryName is the base name of the go-microvm runner process.
+// It must match the constant go-microvm's own terminateStaleRunner
+// keys off so that this cleanup and the upstream one agree on the
+// runner fingerprint.
+const runnerBinaryName = "go-microvm-runner"
+
+// stateLockTimeout bounds how long CleanupStaleVMDirs will wait to
+// acquire a VM state lock before treating the dir as live (a live
+// process holds the lock → owner is alive → skip).
+const stateLockTimeout = 2 * time.Second
 
 // parseSentinel decodes the sentinel file contents. The current format
 // is two lines: PID\nEXEPATH. The previous format was a single PID.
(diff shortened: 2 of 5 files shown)


cmd/bbox/main.go
```diff
@@ -519,9 +519,12 @@ func run(parentCtx context.Context, agentName string, flags runFlags) error {
...
 	infraws.CleanupStaleSnapshots(snapDir, logger)
 
-	// Clean up stale VM log directories from previous crashes.
+	// Clean up stale VM log directories and orphaned data dirs (COW rootfs
+	// clones) left behind by crashed runs.
 	if home, homeErr := os.UserHomeDir(); homeErr == nil {
-		infravm.CleanupStaleLogs(filepath.Join(home, ".config", "broodbox", "vms"), logger)
+		vmsDir := filepath.Join(home, ".config", "broodbox", "vms")
+		infravm.CleanupStaleLogs(vmsDir, logger)
+		infravm.CleanupStaleVMData(vmsDir, logger)
```
JAORMX: **[High — devops-expert] Silently defeats the documented `BBOX_KEEP_VM_DATA=1` debug workflow**

CLAUDE.md (§Debugging & Profiling → "Preserving VM data for debugging") documents `BBOX_KEEP_VM_DATA=1` as the supported way to preserve a crashed VM's data dir for post-mortem (`console.log`, `vm.log`, `rootfs-work/`, `trace.json`). The user's mental model: crash leaves a dir behind → set the env var → re-run → inspect.

`CleanupStaleVMData` runs unconditionally at startup of *every* `bbox` invocation and `os.RemoveAll`s the `data/` subdir of any VM whose state shows `Active==true` with a dead runner PID. A crashed run leaves exactly that — `Active==true`, dead PID. So:

```sh
BBOX_KEEP_VM_DATA=1 bbox claude-code --trace   # crashes
BBOX_KEEP_VM_DATA=1 bbox claude-code --trace   # startup nukes the previous data dir
```

The env var only gates the *current* run's `Stop()` (`internal/infra/vm/runner.go:394`); it does nothing for orphaned dirs from prior runs. This is a correctness finding against a stated requirement (the documented debug procedure), not a style preference.

**Fix:** gate the startup call on the env var being unset:
```go
if os.Getenv("BBOX_KEEP_VM_DATA") != "1" {
    infravm.CleanupStaleVMData(vmsDir, logger)
}
```
`CleanupStaleLogs` has the same hole, but this PR is the moment to not widen it. At minimum, document the interaction in the CLAUDE.md "Preserving VM data" section.

internal/infra/vm/cleanup.go
```diff
@@ -119,6 +121,87 @@ func CleanupStaleLogs(vmsDir string, logger *slog.Logger) {
...
+				"path", dataDir, "error", loadErr)
+			continue
+		}
+
+		// Not an orphan: no active VM recorded, or no runner PID to check.
+		if !st.Active || st.PID <= 0 {
+			continue
+		}
+
+		if process.IsAlive(st.PID) {
```
JAORMX: **[Medium — software-architect] Weaker liveness check than the sibling and than go-microvm's own analogue**

`process.IsAlive(st.PID)` is a bare `kill -0`. The sibling `CleanupStaleLogs` deliberately does better: when the sentinel carries an exe path it uses `process.IsExpectedProcess(pid, exePath)` and documents *why* — "a kill -0 liveness check alone cannot tell [a recycled PID] apart" (`cleanup.go:94-106`). go-microvm's own `terminateStaleRunner` — the exact analogue, operating on the same state file — uses `processIsExpected(st.PID)` against `"go-microvm-runner"` (`go-microvm/microvm.go`), and the state struct even carries `PIDStartTime` (`state/state.go`) specifically to disambiguate recycled PIDs. This function ignores both signals.

The failure mode is a *leak*, not a wrong deletion: a PID recycled onto an unrelated live process is treated as alive and skipped — exactly the case this function exists to reclaim. On a busy, long-lived host a recycled PID can stay alive indefinitely, so "reclaimed on a later run once the PID is free" (function doc) may never arrive.

**Fix:** reuse the fingerprint-aware helper already used by the sibling: `process.IsExpectedProcess(st.PID, "go-microvm-runner")` — the same base-name go-microvm's default `processIsExpected` uses. This keeps the orphan decision consistent with `CleanupStaleLogs` and with go-microvm's own stale-runner check.

The existing test `TestCleanupStaleVMData_KeepsLiveRunner` uses `os.Getpid()` (the test binary), which would no longer match `"go-microvm-runner"`; that test would need to assert the fingerprint-mismatch path explicitly rather than reuse the live-PID shortcut.

cmd/bbox/main.go
```diff
@@ -519,9 +519,12 @@ func run(parentCtx context.Context, agentName string, flags runFlags) error {
 	// Clean up stale snapshot dirs from previous crashes.
 	infraws.CleanupStaleSnapshots(snapDir, logger)
 
-	// Clean up stale VM log directories from previous crashes.
+	// Clean up stale VM log directories and orphaned data dirs (COW rootfs
+	// clones) left behind by crashed runs.
 	if home, homeErr := os.UserHomeDir(); homeErr == nil {
-		infravm.CleanupStaleLogs(filepath.Join(home, ".config", "broodbox", "vms"), logger)
+		vmsDir := filepath.Join(home, ".config", "broodbox", "vms")
+		infravm.CleanupStaleLogs(vmsDir, logger)
```
JAORMX: **[Medium — software-architect] Ordering with `CleanupStaleLogs` is load-bearing in the harmful direction**

`CleanupStaleLogs` runs first, keys off the **bbox** PID in the sentinel, and when dead calls `os.RemoveAll(vmDir)` *recursively* — taking the `data/` subdir with it. `CleanupStaleVMData` keys off the **runner** PID in go-microvm's state file and only removes `data/`. The runner is spawned with `Setsid: true`, so it is in its own session and is **not** killed when bbox is SIGKILLed/OOMed. Therefore the common crash case "bbox died, detached runner + VM still alive" leaves the bbox sentinel PID dead while the runner PID is live.

With the current ordering, the next `bbox` invocation sees a dead bbox sentinel and `os.RemoveAll`s the whole `vmDir` — including `data/rootfs-work/` that the still-running VM is actively serving over virtiofs — *before* `CleanupStaleVMData`'s runner-aware check runs. The runner-aware judge never gets to protect the dir. Consequence: guest-side I/O corruption of a live VM, which is a correctness issue, not just hygiene.

**Fix:** make a vmDir's deletion contingent on *both* owner signals being dead/absent. Either (a) have `CleanupStaleLogs` consult the runner state before its recursive remove (skip / defer to `CleanupStaleVMData` when `data/` holds an active state with a live runner PID), or (b) collapse the two sweeps into one pass per vmDir that reads both the sentinel and the state file and deletes only when neither owner is alive. The destructive recursive delete in `CleanupStaleLogs` is the one that needs the runner-aware guard; the "only-if-empty" `os.Remove(vmDir)` in `CleanupStaleVMData` is already the right non-destructive prune.

**Verification:** a test that constructs a vmDir with (i) a sentinel pointing at a dead bbox PID and (ii) a state file with `Active=true` and a *live* runner PID (use the test process PID), runs both cleanups in order, and asserts the vmDir (and `data/`) survive. Today this would fail.

internal/infra/vm/cleanup.go
```diff
@@ -119,6 +121,87 @@ func CleanupStaleLogs(vmsDir string, logger *slog.Logger) {
...
+			logger.Debug("skipping VM data dir owned by running runner",
+				"path", dataDir, "pid", st.PID)
+			continue
+		}
+
+		logger.Warn("removing orphaned VM data directory",
+			"path", dataDir, "pid", st.PID, "reason", "runner-dead")
+		if err := os.RemoveAll(dataDir); err != nil {
```
JAORMX: **[Medium — devops-expert] Unbounded synchronous `RemoveAll` on a latency-sensitive interactive path**

This runs in `run()` before the VM boots, on every invocation, including the common interactive `bbox claude-code` a human is sitting in front of. `data/` contains the COW `rootfs-work/` clone — hundreds of MB to GB of small files. `os.RemoveAll` is a recursive syscall walk; on a cold cache or busy disk a single orphan can take seconds, and there's no bound on the number of orphans swept in one pass.

For a tool whose whole value prop is "boot a microVM fast" and which already has a `--timings` feature flag for startup-phase latency, doing unbounded synchronous disk reclamation on the critical path is the wrong default. `CleanupStaleLogs` is cheap (small log files); `CleanupStaleVMData` is not, and the two should not share an execution model.

**Fix (pick one, in order of preference):**
1. Move reclamation to an explicit `bbox cleanup` subcommand or a `--cleanup` flag; don't run it on the hot path.
2. Background it: spawn a goroutine (or fork a short-lived child) that does the sweep after the VM is started, so it overlaps with boot rather than serializing before it. Caveat: a background `RemoveAll` racing the *current* run's own dir creation reintroduces the TOCTOU flagged elsewhere — gate by name, not just by liveness.
3. If it must stay synchronous, cap it: at most N orphans per pass, or skip if the dir count exceeds a threshold and emit a single Warn pointing at `bbox cleanup`.

At minimum, add a timing span (`bbox.Prepare` already exists in the trace hierarchy per CLAUDE.md) so latency regressions here are visible in `--timings`/`--trace` output rather than silent.

internal/infra/vm/cleanup.go
```diff
@@ -119,6 +121,87 @@ func CleanupStaleLogs(vmsDir string, logger *slog.Logger) {
...
+			continue
+		}
+
+		vmDir := filepath.Join(vmsDir, entry.Name())
+		dataDir := filepath.Join(vmDir, vmDataSubdir)
+
+		// Load reads go-microvm's state file without locking. A missing
+		// file (no VM ever ran here, or it was already cleaned) yields a
+		// default inactive state, which we skip below.
+		st, loadErr := state.NewManager(dataDir).Load()
```
JAORMX: **[Low — secure-code-reviewer] No size cap on state-file read (CWE-400)**

`state.NewManager(dataDir).Load()` delegates to the dependency's `loadState` → `os.ReadFile` (unbounded). A same-UID attacker (or a corrupted disk/faulty runner) that plants a multi-GiB `go-microvm-state.json` in an orphaned `data/` dir causes a memory spike / OOM at bbox startup.

Notably, the *sibling* `CleanupStaleLogs` path in this same file was deliberately hardened with `maxSentinelSize = 4096` and `readSentinelFile` for exactly this threat (comment at `cleanup.go:22-27`); the new function does not apply the same discipline to its on-disk input.

Same-UID, so this is hardening rather than a trust-boundary crossing — hence Low. A state file legitimately fits in well under 4 KiB. **Fix:** cap the read at the call site (read with `io.LimitReader` and reject oversize, mirroring `readSentinelFile`) or upstream a size guard into `state.Manager.loadState`.

internal/infra/vm/cleanup.go
```diff
@@ -119,6 +121,87 @@ func CleanupStaleLogs(vmsDir string, logger *slog.Logger) {
 	}
 }
 
+// vmDataSubdir is the per-VM data subdirectory (under ~/.config/broodbox/vms/<name>/)
+// that holds the go-microvm state file and the COW rootfs clone (rootfs-work/).
+const vmDataSubdir = "data"
```
JAORMX: **[Low — software-architect] `vmDataSubdir = "data"` duplicated by a string literal in `vmDataDir`**

The path convention `<vmsDir>/<name>/data` is authored in two places: `vmDataDir()` (`internal/infra/vm/runner.go:441`, `filepath.Join(home, ".config", "broodbox", "vms", name, "data")`) and this new `vmDataSubdir` const (used by the cleanup that scans for that dir). The const exists precisely to name the convention; `vmDataDir` ignores it and hardcodes the literal. If one changes, the cleanup silently scans the wrong subdir and reclaims nothing — a latent desync with no compile-time signal.

**Fix:** define the const once next to `vmDataDir` (the path-authoritative function) and use it in both `vmDataDir` and `CleanupStaleVMData`. The const belongs in brood-box, *not* alongside go-microvm's path conventions — go-microvm's `WithDataDir` takes an opaque path; the `data` segment is a brood-box layout choice.

internal/infra/vm/cleanup.go
JAORMX: **[Medium — cross-confirmed: secure-code-reviewer + devops-expert + software-architect] TOCTOU between unlocked state load and recursive delete**

Three reviewers independently flagged the window between `state.NewManager(dataDir).Load()` (no lock) and `os.RemoveAll(dataDir)`. The runner is spawned with `Setsid: true`, so the "bbox died, detached runner + VM still alive" crash case leaves a live runner PID in a dir whose bbox sentinel PID is dead.

`state.Load()` is documented as "useful for read-only access where consistency with subsequent writes is not needed." `Save()` is atomic (temp + rename), so a torn read isn't the risk — the decision window is. A concurrent `bbox` starting in the same uniquely-named dir could be mid-setup when this process reads a stale dead-PID state and removes the dir out from under it.

The sibling `CleanupStaleLogs` does *not* have this problem because it keys on the sentinel file the live process owns, not on state a live process is concurrently writing.

**Fix:** use `LoadAndLock(ctx)` (or `LoadAndLockWithRetry` with a short timeout) so cleanup and a starting VM serialize on the same flock rather than both touching the dir unlocked:

```go
ls, err := state.NewManager(dataDir).LoadAndLock(ctx)
if err != nil {
    logger.Debug("skipping VM dir with unreadable/locked state", "path", dataDir, "error", err)
    continue
}
st := ls.State
defer func() { _ = ls.Release() }()
// ... decide + RemoveAll while still holding the lock ...
```

CWE-367 / CWE-663.

JAORMX (commented):
## Panel review — 3 axes, 5 reviewers

Ran a three-axis panel review (Spec / Standards / Domain) with the Domain axis fanned out to `secure-code-reviewer`, `software-architect`, `devops-expert`, and `library-reuse-reviewer`. Findings are inline as line-anchored comments; summary below. (Submitting as COMMENT rather than REQUEST_CHANGES since GitHub doesn't allow requesting changes on your own PR — but please treat the High finding as a merge blocker.)

### Spec — ✅ clean (with one scope-creep note)
All core requirements in issue #87 are satisfied (scan `~/.config/broodbox/vms/*/data/`, check `active: true` + dead PID, remove orphaned data dirs, wire into `internal/infra/vm/` called from `cmd/bbox/main.go`). The spec said `state.json`; the PR correctly uses `state.Manager.Load()` which reads `go-microvm-state.json` (the actual go-microvm filename) — an acceptable correction, not a mismatch. One scope-creep note: the `os.Remove(vmDir)` parent-prune is defensible (reclaims `--log-file` runs) but wasn't asked for; left as-is since it's the right call.

### Standards — ✅ clean
SPDX headers present, `log/slog` exclusively, correct DDD layer (`internal/infra/vm/`, no domain/sandbox/client leakage, `go-microvm/state` import is consistent — `internal/infra/vm/` is the go-microvm adapter), table-driven tests alongside code, imperative commit message.

### Domain — 1 High ship-blocker, 4 Medium, 2 Low

**Ship-blockers (1)**
- **[High]** `CleanupStaleVMData` runs unconditionally at startup and silently defeats the documented `BBOX_KEEP_VM_DATA=1` debug workflow — a crashed run's `data/` dir is exactly what the env var is meant to preserve, and this cleanup deletes it before the user can inspect. `cmd/bbox/main.go:527`.

**Cross-confirmed (1)**
- **[Medium]** TOCTOU between unlocked `state.Load()` and `os.RemoveAll` — flagged independently by secure-code-reviewer, devops-expert, and software-architect. A concurrent `bbox` starting in the same dir can have its rootfs clone deleted out from under it. Use `LoadAndLock(ctx)`. CWE-367/CWE-663. `internal/infra/vm/cleanup.go:167`.

**Mechanical fixes (3)**
- **[Medium]** Weaker liveness check than the sibling and than go-microvm's own analogue — `process.IsAlive(st.PID)` (bare `kill -0`) ignores the `PIDStartTime` / exe-path fingerprint that `CleanupStaleLogs` and go-microvm's `terminateStaleRunner` both use. Recycled-PID orphans leak indefinitely. Use `process.IsExpectedProcess(st.PID, "go-microvm-runner")`. `cleanup.go:179`.
- **[Medium]** Ordering with `CleanupStaleLogs` is load-bearing in the harmful direction — `CleanupStaleLogs` runs first and `os.RemoveAll`s the whole `vmDir` (incl. `data/rootfs-work/`) on a dead bbox sentinel, *before* this function's runner-aware check can protect a still-live detached runner. Guest-side I/O corruption of a live VM. `main.go:526`.
- **[Medium]** Unbounded synchronous `RemoveAll` on a latency-sensitive interactive path — hundreds of MB of small files removed on every `bbox` invocation before the VM boots. Move to `bbox cleanup` subcommand, background it, or cap it. `cleanup.go:187`.

**Polish (2)**
- **[Low]** No size cap on the `state.Load()` read (CWE-400) — the sibling `CleanupStaleLogs` caps its sentinel read at 4 KiB; this function reads an unbounded `os.ReadFile`. `cleanup.go:167`.
- **[Low]** `vmDataSubdir = "data"` const is duplicated by a string literal in `vmDataDir()` (`runner.go:441`) — latent desync with no compile-time signal. `cleanup.go:126`.

### Gaps
- `library-reuse-reviewer` returned clean — no reinventions; the scan-loop shape is incidental duplication (Rule of Three not met), `IsAlive` is the right primitive here (no exe path in state to fingerprint against — though see the Medium on using the runner binary name).

**Most important single issue:** the High `BBOX_KEEP_VM_DATA` defeat — it directly breaks a documented debugging procedure. The TOCTOU (cross-confirmed Medium) is the highest-confidence finding since three reviewers independently reached it.

cmd/bbox/main.go
JAORMX: Addressed in 11ff412. The startup sweep is now gated on `BBOX_KEEP_VM_DATA != "1"` (a debug line is logged on skip), and CLAUDE.md "Preserving VM data for debugging" documents the dual gating — the env var preserves the *current* run's `Stop()` **and** disables startup stale-dir reclamation, so a crashed prior run (`Active==true` + dead PID) survives for re-run-and-inspect.

internal/infra/vm/cleanup.go
JAORMX: Addressed in 11ff412. The runner PID is now fingerprinted via `process.IsExpectedProcess(st.PID, "go-microvm-runner")` — the same base name go-microvm's own `terminateStaleRunner` keys off (new `runnerBinaryName` const in `cleanup.go`). `runnerAlive = st.Active && st.PID > 0 && fingerprint match`. The old `TestCleanupStaleVMData_KeepsLiveRunner` shortcut is gone; `TestCleanupStaleVMDirs_FingerprintMismatchRemoved` (Linux-only, darwin-skipped with a message) asserts the recycled-PID path explicitly, and `TestCleanupStaleVMDirs` `active_state_dead_pid_removed_all_platforms` covers the cross-platform dead-runner reclaim.

cmd/bbox/main.go
JAORMX: Addressed in 11ff412 via your option (b): `CleanupStaleLogs` + `CleanupStaleVMData` are collapsed into one `CleanupStaleVMDirs` that does a single pass per vmDir, checking both the bbox sentinel and the runner state and reclaiming only when *neither* owner is alive. The two-phase "logs-first recursive RemoveAll, then runner-aware check" ordering hole is gone. `TestCleanupStaleVMDirs_FingerprintMismatchRemoved` constructs exactly the case you described (dead bbox sentinel + live runner PID via `os.Getpid()`) and asserts the vmDir survives — on Linux, where the fingerprint mismatch is detectable; darwin skips since `IsExpectedProcess` falls back to `IsAlive` and the live PID is correctly treated as alive → preserved.

internal/infra/vm/cleanup.go
JAORMX: Addressed in 11ff412 with your option 2 (background). The sweep now runs in a goroutine at the `cmd/bbox/main.go` call site, overlapping reclamation with VM boot rather than serializing before it. The current run's `vmName` is passed as `skipName` so there's no TOCTOU against the current run's own dir creation. Start and elapsed duration are logged at debug — that gives the latency visibility you asked for without a new OTel span (no tracer is in scope at that early point in `run()`). No new `bbox cleanup` subcommand was added (out of scope for this PR).

internal/infra/vm/cleanup.go
JAORMX: Addressed in 11ff412. Added `const maxStateSize = 4096` next to `maxSentinelSize` and a `stateFileOversize(dataDir)` helper that walks `os.ReadDir(dataDir)` and returns true if any regular file exceeds the cap — called before `LoadAndLock` so the unbounded `os.ReadFile` underneath never runs on a planted giant. Mirrors the sentinel discipline; same-UID hardening. `TestCleanupStaleVMDirs` `oversized_state_file_preserved` and a direct `TestStateFileOversize_Direct` table (under/over/exactly-at-limit, nonexistent dir) cover it.

internal/infra/vm/cleanup.go
JAORMX: Addressed in 11ff412. `vmDataSubdir = "data"` moved to `runner.go` next to `vmDataDir` (the path-authoritative function) and is now used in its `filepath.Join`; the duplicate const in `cleanup.go` is gone. Doc comment notes the `data` segment is a brood-box layout choice — go-microvm's `WithDataDir` takes an opaque path — so the const belongs here, not alongside go-microvm's path conventions.

internal/infra/vm/cleanup.go
JAORMX: Addressed in 11ff412 — the cross-confirmed TOCTOU. The sweep now uses `state.NewManager(dataDir).LoadAndLockWithRetry(ctx, 2*time.Second)`; the flock is held across the decision **and** the `RemoveAll`, then `Release()`d. `os.Stat(dataDir)` first avoids `LoadAndLock`'s `MkdirAll` side effect creating dirs that don't exist. On lock-acquisition failure (a live process holds it) the dir is treated as live and skipped — that's the direct closure of the "concurrent bbox starting in the same dir" window. `TestCleanupStaleVMDirs_HeldLockSkipped` holds the lock from another goroutine via `LoadAndLock` and asserts the dir survives the sweep. CWE-367/CWE-663.

PR #199, opened 2026-06-26
Prune stale firmware cache versions after download

### LINKED ISSUE #12: Prune stale firmware cache versions
## Problem

Each propolis version bump creates a new firmware cache directory at `~/.cache/broodbox/firmware/<version>/<os>-<arch>/` (~tens of MBs per version). Old version directories are never deleted, so the cache grows unbounded over time.

The cache structure looks like:

```
~/.cache/broodbox/firmware/
├── v0.0.7/
│   └── linux-amd64/
│       ├── firmware.json
│       └── propolis-firmware-linux-amd64/
│           └── libkrunfw.so.5
├── v0.0.8/                          ← current
│   └── linux-amd64/
│       ├── firmware.json
│       └── propolis-firmware-linux-amd64/
│           └── libkrunfw.so.5
└── .firmware.lock
```

Only the current version is ever used. Previous versions are dead weight.

## Proposed Solution

After a successful fresh firmware download in `downloadFirmware`, scan sibling version directories under the cache root and delete any that don't match the version just downloaded.

### Implementation

In `internal/infra/vm/firmware.go`, after writing the manifest and before returning from the successful download path:

1. List entries in `cacheRoot` (the firmware cache root, e.g. `~/.cache/broodbox/firmware/`)
2. For each subdirectory that looks like a version (skip `.firmware.lock` and temp files)
3. If the directory name doesn't match the current `version`, remove it
4. Log a debug message for each pruned version
5. Pruning failures should be logged but not cause the download to fail — firmware is already cached successfully at this point

### Scope

- Only prune after a **fresh download** (not on cache hit — no point scanning when nothing changed)
- Only prune **version directories**, not the lock file or temp files
- The file lock is already held during download, so no concurrency issues
- Should be safe to also add a `task firmware-clean` target that removes the entire `~/.cache/broodbox/firmware/` directory, for manual cleanup

### Not in scope

- LRU/TTL-based eviction (overkill — only one version is ever active)
- Pruning runtime cache (separate concern)

JAORMX:
## Summary

Fixes #12 — the firmware cache grew unbounded across go-microvm version bumps.

Each version bump creates a new `~/.cache/broodbox/firmware/<version>/` directory (~tens of MB), but only the version pinned in `go.mod` is ever used. Old version dirs were never deleted.

## Approach

In `downloadFirmware`, after a **successful fresh download** (manifest written, firmware lock still held), call `pruneStaleFirmwareVersions`:

- Scan sibling entries under the cache root.
- Remove any **directory** whose name isn't the version just downloaded.
- Skip transient/temp entries: `.firmware.lock` (a file), `firmware-*.tar.gz` temp archives (files), and `firmware-extract-*` temp dirs (`firmware-` prefix guard).
- Runs **only on a fresh download**, never on a cache hit (nothing changed, no point scanning).
- Failures are logged at debug level and **never propagated** — the firmware is already cached successfully, so pruning is strictly best-effort.

The file lock is held throughout, so there are no concurrency concerns.

Also adds a `task firmware-clean` target to wipe the whole firmware cache manually (honors `XDG_CACHE_HOME`), as suggested in the issue.

## Tests

- multiple stale versions removed; current version, lock file, and transient temp entries all preserved
- no-op when only the current version is present
- no panic when the cache root doesn't exist

Verified: `task lint` (0 issues), full `task test` passes (incl. `-race`).

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -236,6 +236,11 @@ tasks:
       - rm -f internal/infra/vm/runtimebin/VERSION internal/infra/vm/runtimebin/LICENSE-GPL
       - rm -f internal/infra/vm/runtimebin/sha256sums.txt
 
+  firmware-clean:
+    desc: Remove the entire firmware download cache (~/.cache/broodbox/firmware)
+    cmds:
+      - rm -rf "${XDG_CACHE_HOME:-$HOME/.cache}/broodbox/firmware"
+
   # --- OCI guest images ---
 
   image-base:
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -237,9 +237,20 @@ tasks:
       - rm -f internal/infra/vm/runtimebin/sha256sums.txt
 
   firmware-clean:
-    desc: Remove the entire firmware download cache (~/.cache/broodbox/firmware)
+    desc: Remove the entire firmware download cache
     cmds:
-      - rm -rf "${XDG_CACHE_HOME:-$HOME/.cache}/broodbox/firmware"
+      - rm -rf "{{.BBOX_CACHE_HOME}}/broodbox/firmware"
+    vars:
+      BBOX_CACHE_HOME:
+        sh: |
+          base="${XDG_CACHE_HOME:-}"
+          if [ -z "$base" ]; then
+            case "$(uname -s)" in
+              Darwin) base="$HOME/Library/Caches" ;;
+              *)      base="$HOME/.cache" ;;
+            esac
+          fi
+          echo "$base"
 
   # --- OCI guest images ---
 
--- a/internal/infra/vm/firmware.go
+++ b/internal/infra/vm/firmware.go
@@ -276,6 +276,11 @@ func downloadFirmware(ctx context.Context, cacheRoot, version, osName, arch stri
 			return FirmwareResolution{}, err
 		}
 
+		// Fresh download succeeded — reclaim cache dirs left by older
+		// firmware versions. Only the version pinned in go.mod is ever
+		// used, so previous version dirs (tens of MB each) are dead weight.
+		pruneStaleFirmwareVersions(ctx, cacheRoot, version)
+
 		slog.DebugContext(ctx, "firmware downloaded", "dir", filepath.Dir(finalFwPath), "version", version, "arch", candidate)
 		return FirmwareResolution{
 			Dir:       filepath.Dir(finalFwPath),
@@ -292,6 +297,49 @@ func downloadFirmware(ctx context.Context, cacheRoot, version, osName, arch stri
 		lastErr = errors.New("firmware download failed")
 	}
 	return FirmwareResolution{}, lastErr
+}
+
+// pruneStaleFirmwareVersions removes firmware cache version directories under
+// cacheRoot that don't match keepVersion. Each go-microvm version bump creates
+// a new <cacheRoot>/<version>/ directory, but only the pinned version is ever
+// used, so older ones accumulate unbounded.
+//
+// It must be called only after a successful fresh download, with the firmware
+// lock held (no concurrent writers). Failures are logged at debug level and
+// never propagated: the firmware is already cached successfully, so pruning is
+// strictly best-effort cleanup. The lock file and transient temp entries
+// (firmware-*.tar.gz archives, firmware-extract-* dirs) are left untouched.
+func pruneStaleFirmwareVersions(ctx context.Context, cacheRoot, keepVersion string) {
+	entries, err := os.ReadDir(cacheRoot)
+	if err != nil {
+		slog.DebugContext(ctx, "firmware prune: cannot scan cache root", "root", cacheRoot, "error", err)
+		return
+	}
+
+	for _, entry := range entries {
+		// Only version directories are pruned; this skips .firmware.lock
+		// and any leftover firmware-*.tar.gz temp archives (plain files).
+		if !entry.IsDir() {
+			continue
+		}
+		name := entry.Name()
+		if name == keepVersion {
+			continue
+		}
+		// Skip transient temp dirs created in cacheRoot during download.
+		if strings.HasPrefix(name, "firmware-") {
+			continue
+		}
+
+		stalePath := filepath.Join(cacheRoot, name)
+		if err := os.RemoveAll(stalePath); err != nil {
+			slog.DebugContext(ctx, "firmware prune: failed to remove stale version",
+				"path", stalePath, "error", err)
+			continue
+		}
+		slog.DebugContext(ctx, "firmware prune: removed stale version",
+			"path", stalePath, "version", name)
+	}
 }
 
 func findSystemFirmware(version, osName, arch string) (FirmwareResolution, error) {
--- a/internal/infra/vm/firmware.go
+++ b/internal/infra/vm/firmware.go
@@ -34,6 +34,13 @@ const (
 	maxFirmwareExtractSize = 128 << 20
 	// maxFirmwareEntries caps the number of tar entries to prevent inode exhaustion.
 	maxFirmwareEntries = 1000
+	// firmwareTempPrefix is the shared prefix of transient download entries
+	// created directly under cacheRoot (firmware-*.tar.gz archives and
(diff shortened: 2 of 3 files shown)


Taskfile.yaml
```diff
@@ -236,6 +236,11 @@ tasks:
       - rm -f internal/infra/vm/runtimebin/VERSION internal/infra/vm/runtimebin/LICENSE-GPL
       - rm -f internal/infra/vm/runtimebin/sha256sums.txt
 
+  firmware-clean:
+    desc: Remove the entire firmware download cache (~/.cache/broodbox/firmware)
+    cmds:
+      - rm -rf "${XDG_CACHE_HOME:-$HOME/.cache}/broodbox/firmware"
```
JAORMX: **[Medium — devops-expert] macOS cache-root drift**

On macOS (an explicitly supported platform), the Go code resolves the cache via `github.com/adrg/xdg`'s `xdg.CacheHome`, which defaults to `$HOME/Library/Caches` — not `$HOME/.cache`. This Taskfile target hardcodes `${XDG_CACHE_HOME:-$HOME/.cache}`, so on macOS with `XDG_CACHE_HOME` unset (the normal case), `task firmware-clean` silently removes the wrong path: `bbox` writes firmware to `~/Library/Caches/broodbox/firmware/` but this target removes `~/.cache/broodbox/firmware/`. The stale firmware keeps accumulating and the operator gets a false success.

The rest of the codebase already has this split: `imageCacheDir()` owns the image cache path (not the Taskfile). The cleanest fix is to delegate to the binary (`bin/bbox cache firmware purge`, mirroring `cmd/bbox/cache.go`) so there's a single source of truth. If keeping it in the Taskfile, mirror `xdg.CacheHome`'s resolution:

```yaml
firmware-clean:
  desc: Remove the entire firmware download cache
  cmds:
    - rm -rf "{{.BBOX_CACHE_HOME}}/broodbox/firmware"
vars:
  BBOX_CACHE_HOME:
    sh: |
      base="${XDG_CACHE_HOME:-}"
      if [ -z "$base" ]; then
        case "$(uname -s)" in
          Darwin) base="$HOME/Library/Caches" ;;
          *)      base="$HOME/.cache" ;;
        esac
      fi
      echo "$base"
```

(Pre-existing doc drift in `README.md:94` and `docs/MACOS.md:37` also point at `~/.cache/...` — out of scope here but worth a follow-up.)

internal/infra/vm/firmware.go
```diff
@@ -294,6 +299,49 @@ func downloadFirmware(ctx context.Context, cacheRoot, version, osName, arch stri
...
+		// and any leftover firmware-*.tar.gz temp archives (plain files).
+		if !entry.IsDir() {
+			continue
+		}
+		name := entry.Name()
+		if name == keepVersion {
+			continue
+		}
+		// Skip transient temp dirs created in cacheRoot during download.
+		if strings.HasPrefix(name, "firmware-") {
```
JAORMX: **[Info — cross-confirmed: software-architect + library-reuse-reviewer] Magic-string coupling to download temp prefixes**

The `firmware-` prefix guard mirrors `os.CreateTemp(cacheRoot, "firmware-*.tar.gz")` and `os.MkdirTemp(cacheRoot, "firmware-extract-")` in `downloadFirmware`. If either temp prefix ever changes, prune silently stops protecting transient entries.

Two reviewers independently flagged this, with the same conclusion: it's defense-in-depth, not a load-bearing invariant. By prune time this process's own temps are already cleaned up and the `flock` excludes concurrent writers, so the only `firmware-`-prefixed dirs that can exist here are leftovers from a *crashed* prior run — which are garbage, not valid artifacts (the rename at finalize moves `tmpDir` → version-scoped `cacheDir`, so no `firmware-`-prefixed path is ever permanent). Failure mode of drift is "leaves garbage," not "deletes something needed," so it won't rot dangerously.

Optional hardening: hoist the prefix into a shared constant both sites reference. Not worth blocking on.

internal/infra/vm/firmware.go
```diff
@@ -294,6 +299,49 @@ func downloadFirmware(ctx context.Context, cacheRoot, version, osName, arch stri
...
+			continue
+		}
+
+		stalePath := filepath.Join(cacheRoot, name)
+		if err := os.RemoveAll(stalePath); err != nil {
+			slog.DebugContext(ctx, "firmware prune: failed to remove stale version",
+				"path", stalePath, "error", err)
```
JAORMX: **[Info — secure-code-reviewer] Consider `Warn` for unexpected `RemoveAll` errors**

Defensible as best-effort cleanup, but a consistently-failing `os.RemoveAll` (e.g. permissions drift, read-only mount) logged only at `Debug` will silently fill the cache disk over many version bumps — the exact problem this PR is solving. Consider reserving `Debug` for "not exist"/"cannot scan" (legitimately quiet) and bumping unexpected `RemoveAll` failures to `Warn` so they're visible in default log output. Operational hardening, not a correctness/security issue — no action required to merge.

JAORMX (commented):
## Panel review — 3 axes, 4 reviewers

Ran a three-axis panel review (Spec / Standards / Domain) with the Domain axis fanned out to `secure-code-reviewer`, `software-architect`, `devops-expert`, and `library-reuse-reviewer`. Findings are inline as line-anchored comments; summary below.

### Spec — ✅ clean
All five implementation requirements in issue #12 are satisfied (list `cacheRoot`, skip lock/temp entries, remove non-current version dirs, debug log per prune, failures don't fail the download). No scope creep — the `task firmware-clean` target is explicitly in scope. Call site is fresh-download-only with the `flock` held, matching the spec.

### Standards — ✅ clean
SPDX headers present, `log/slog` exclusively, correct DDD layer (`internal/infra/vm/`, no domain/sandbox/client leakage), table-driven tests alongside code, imperative commit message.

### Domain — 0 ship-blockers, 1 mechanical, 1 cross-confirmed Info, 1 polish

- **[Medium]** `firmware-clean` resolves the cache root differently from the Go code on macOS → silently no-ops on a supported platform. See Taskfile.yaml comment.
- **[Info — cross-confirmed]** `firmware-` prefix guard is magic-string coupled to the download temp prefixes. Two reviewers independently concluded it's defense-in-depth, not load-bearing (failure mode is "leaves garbage," not "deletes something needed"). Optional shared constant.
- **[Info]** Prune failures at `Debug` may hide a consistently-failing prune from default log output. Consider `Warn` for unexpected `RemoveAll` errors.

**Most important single issue:** the Medium macOS path drift in `firmware-clean` — the only finding that changes actual behavior on a supported platform. The Go-side change (`pruneStaleFirmwareVersions`) is sound; secure-code-reviewer verified no symlink/TOCTOU/path-traversal exposure under the actual threat model (0700 owner-only cache root, lock held, stdlib `RemoveAll` doesn't follow symlinks).

Taskfile.yaml
JAORMX: Addressed in 080cb92. The `firmware-clean` target now resolves the cache root the same way the Go code does (`xdg.CacheHome`): `XDG_CACHE_HOME` first, then `Darwin → $HOME/Library/Caches`, else `$HOME/.cache`. So on macOS with `XDG_CACHE_HOME` unset it now removes `~/Library/Caches/broodbox/firmware/` — the path `bbox` actually writes to — instead of silently clearing the wrong location.

The README/MACOS doc drift (`~/.cache/...`) is noted as out of scope here; happy to follow up separately.

internal/infra/vm/firmware.go
JAORMX: Addressed in 080cb92. Unexpected `os.RemoveAll` failures now log at `Warn`; only the `os.ErrNotExist` case stays at `Debug` (legitimately quiet). So a persistently-failing removal (permissions drift, read-only mount) surfaces in default log output instead of silently refilling the cache disk over version bumps.

internal/infra/vm/firmware.go
JAORMX: Addressed in 080cb92. Hoisted the prefix into a shared `firmwareTempPrefix = "firmware-"` constant, now referenced by both `os.CreateTemp` / `os.MkdirTemp` patterns and the prune skip-guard, so the coupling is explicit and drift can't silently disable the transient-entry guard.

Issue #191, opened 2026-06-18
Add bring-your-own AI agent support


JAORMX:
## Summary

Add first-class support for **bring your own AI agent/client/harness** in Brood Box, so users can run and register arbitrary AI coding tools without forking this repo, adding a `pkg/clients/<name>` package, or maintaining an upstream Dockerfile here.

The north-star UX should be:

```bash
bbox agents import ghcr.io/acme/aider-bbox:latest
bbox agents doctor aider
bbox aider
```

And for simple/power-user cases:

```yaml
agents:
  my-harness:
    image: ghcr.io/me/my-harness:latest
    command: ["my-harness"]
    env_forward: ["MY_HARNESS_*", "OPENAI_API_KEY"]
    mcp:
      mode: env
```

No Go code, no Brood Box fork, no custom infra package, and no built-in Dockerfile should be required for the common path.

## Background / current state

Brood Box already has the right internal seam:

- `pkg/domain/agent.Agent` is the declarative runtime spec: image, command, env forwarding, defaults, egress hosts, credential paths, settings manifest.
- `pkg/domain/agent.ClientEntry` pairs an `Agent` with an optional `Plugin`.
- `pkg/domain/agent.Plugin` currently supports behavior that cannot be expressed declaratively, especially:
  - `MCPConfig() agent.MCPInjector`
  - `Seeder() credential.Seeder`
- Built-ins live under `pkg/clients/<name>` and expose `New() agent.ClientEntry`.
- The in-memory registry supports both:
  - data-only agents via `Registry.Add(agent.Agent)`
  - full client entries via `Registry.AddEntry(agent.ClientEntry)`.
- Global config currently supports unknown/custom agents through `agents.<name>` but only for a limited data-only subset: image, command, env, CPUs, memory, etc.

The pain point: each first-class harness currently needs custom code and often custom image wiring in this repo:

- `pkg/clients/claudecode`
- `pkg/clients/codex`
- `pkg/clients/gemini`
- `pkg/clients/hermes`
- `pkg/clients/opencode`
- `images/<name>/Dockerfile`
- Taskfile image targets

That does not scale to arbitrary agents like Aider, Goose, Cursor Agent CLI, custom internal harnesses, research agents, etc.

## Design principle

Treat this as two layers:

1. **Declarative BYO agents** for the common CLI path.
2. **Trusted behavioral clients** for built-ins and SDK consumers.

[...]

PR #200, opened 2026-06-26
Add declarative bring-your-own agent support (#191 Phase 1)

JAORMX:
## Summary

Implements **Phase 1 of #191** — run arbitrary coding agents from global config with **no Go code, no fork, and no Dockerfile in this repo**. This is the issue's "first useful iteration" acceptance criteria:

```yaml
agents:
  aider:
    image: ghcr.io/acme/aider-bbox:latest
    command: ["aider"]
    env_forward: [OPENAI_API_KEY, "AIDER_*"]
    mcp:
      mode: env
```

```console
$ bbox agents doctor aider
[PASS] config validation: name, image, command, paths, egress all valid
[PASS] required env OPENAI_API_KEY present
$ bbox aider     # boots the image, runs the command
```

## What's included

- **Extended `AgentOverride`** (`pkg/domain/config`): `description`, `default_env`, `env_required`, `credentials.persist`, `settings` entries, per-profile `egress_hosts`, `mcp.mode`.
- **Pure mapping + validation**: `AgentFromOverride` (config→`agent.Agent`) and `ValidateCustomAgent` are pure, table-tested domain functions. The image-ref parser is injected as a closure from the composition root, keeping `go-containerregistry` out of the domain layer.
- **Universal `BBOX_*` env** (`agent.BuildUniversalEnv`) injected into *every* VM: `BBOX_AGENT_NAME`, `BBOX_WORKSPACE`, `BBOX_HOME`, `BBOX_SESSION_ID`, `BBOX_GIT_TOKEN_AVAILABLE`, `BBOX_SSH_AGENT_AVAILABLE`, plus `BBOX_MCP_URL`/`BBOX_MCP_AUTHZ_PROFILE` when MCP is active. Applied **after** forwarded host vars so it's authoritative. `BBOX_MCP_URL` uses a new shared `config.MCPEndpointPath` (`/mcp`) constant — de-duplicating the path previously hardcoded in the proxy and all five built-in clients.
- **`mcp.mode: env`**: enables the proxy without a config-file injector; the agent discovers it via `BBOX_MCP_URL`. (`mcp.mode: config` is rejected with a clear message.)
- **`bbox agents list | inspect | doctor`**: `inspect` shows field **provenance** (built-in/global/workspace) and env **names only, never values**; `doctor` exits non-zero on invalid config or missing required env. `bbox list` remains a shared alias.

## Defaults & security

- **Safer-than-built-in defaults** for custom agents: empty `env_forward` (forward nothing), egress profile `standard` (must declare hosts or set `permissive`), MCP authz `safe-tools`.
- **Workspace-local `.broodbox.yaml` is tighten-only**: it can never add a custom agent or new credential paths, repoint an existing agent's image/command, or widen `env_forward` (subset-coverage enforcement prevents secret exfiltration). Credential/settings paths are validated relative and cannot escape the home dir.
- A **guard test** pins go-microvm's `topology.GatewayIP` to the literal `pkg/sandbox` uses for `BBOX_MCP_URL`, catching drift.

## Out of scope (deliberate follow-ups)

`agents import`/`export`, OCI-embedded manifests, `mcp.mode: config`, `agents add`/`init` scaffolding, and trusted host-side plugins.

## Verification

- `task fmt` clean, `task lint` **0 issues**, full `task test` (race) **passes**.
- End-to-end smoke-tested: `agents list` shows custom agents; `inspect` renders provenance + `MISSING` required-env markers without leaking values; `doctor` passes with required env set (exit 0), fails without it (exit 1), and rejects `mcp.mode: config` (exit 1); `bbox list` unchanged.
- Strict DDD layering preserved (pure domain mapping/validation; I/O only in infra/cmd).

## How it was built

Produced via a multi-agent orchestration loop: **architect → implement → 5-axis panel review (architecture / security / correctness / api-ux / tests) → fix**, iterated 3 rounds. All findings were resolved; the final state was independently re-verified and reviewed before this PR.

### DIFF
--- a/cmd/bbox/agents.go
+++ b/cmd/bbox/agents.go
@@ -1,0 +1,466 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stacklok/brood-box/pkg/clients"
+	domainagent "github.com/stacklok/brood-box/pkg/domain/agent"
+	domainconfig "github.com/stacklok/brood-box/pkg/domain/config"
+	"github.com/stacklok/brood-box/pkg/domain/egress"
+	"github.com/stacklok/brood-box/pkg/domain/settings"
+)
+
+// agentsCmd is the `bbox agents` command group: list, inspect, doctor.
+func agentsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "agents",
+		Short: "Inspect and validate available agents",
+	}
+	cmd.AddCommand(agentsListCmd())
+	cmd.AddCommand(agentsInspectCmd())
+	cmd.AddCommand(agentsDoctorCmd())
+	return cmd
+}
+
+func agentsListCmd() *cobra.Command {
+	var cfgPath string
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List available agents (built-in and custom)",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			return runAgentsList(cmd, cfgPath)
+		},
+	}
+	cmd.Flags().StringVar(&cfgPath, "config", "", "Config file path (default: ~/.config/broodbox/config.yaml)")
+	return cmd
+}
+
+// runAgentsList resolves the registry (built-ins + custom agents from config)
+// and prints each agent's name and image. Shared by `bbox list` and
+// `bbox agents list`.
+func runAgentsList(cmd *cobra.Command, cfgPath string) error {
+	ws, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("getting current directory: %w", err)
+	}
+	resolved, err := buildResolvedRegistry(cfgPath, ws, slog.Default(), io.Discard)
+	if err != nil {
+		return err
+	}
+	out := cmd.OutOrStdout()
+	for _, e := range resolved.registry.List() {
+		_, _ = fmt.Fprintf(out, "%-15s %s\n", e.Agent.Name, e.Agent.Image)
+	}
+	return nil
+}
+
+func agentsInspectCmd() *cobra.Command {
+	var cfgPath string
+	cmd := &cobra.Command{
+		Use:   "inspect <name>",
+		Short: "Show an agent's resolved configuration and field sources",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runAgentsInspect(cmd, cfgPath, args[0])
+		},
+	}
+	cmd.Flags().StringVar(&cfgPath, "config", "", "Config file path (default: ~/.config/broodbox/config.yaml)")
+	return cmd
+}
+
+func runAgentsInspect(cmd *cobra.Command, cfgPath, agentName string) error {
+	ws, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("getting current directory: %w", err)
+	}
+	resolved, err := buildResolvedRegistry(cfgPath, ws, slog.Default(), io.Discard)
+	if err != nil {
+		return err
+	}
+
+	entry, err := resolved.registry.Get(agentName)
+	if err != nil {
+		// Not registered. If it is a custom agent declared in the merged config
+		// (skipped during registration because it failed validation), build a
+		// best-effort view from the override so the user can still see what they
+		// declared. Direct them to `agents doctor` for the failure detail.
+		override, declared := resolved.merged.Agents[agentName]
+		if !declared {
+			return fmt.Errorf("unknown agent %q", agentName)
+		}
+		ag, mapErr := domainconfig.AgentFromOverride(agentName, override, resolved.merged.Defaults)
+		if mapErr != nil {
+			return fmt.Errorf("agent %q is misconfigured: %w (run 'bbox agents doctor %s' for details)", agentName, mapErr, agentName)
+		}
+		out := cmd.OutOrStdout()
+		_, _ = fmt.Fprintf(out, "Warning: agent %q is not registered (failed validation) — run 'bbox agents doctor %s' for details\n\n", agentName, agentName)
+		renderInspect(out, ag, resolved, agentName)
+		return nil
+	}
+
+	out := cmd.OutOrStdout()
+	renderInspect(out, entry.Agent, resolved, agentName)
+	return nil
+}
+
+// renderInspect prints the resolved agent fields with provenance. Security:
+// environment VALUES are never printed — only variable names/patterns and a
+// present/missing indicator for required vars.
+func renderInspect(w io.Writer, ag domainagent.Agent, resolved *resolvedRegistry, agentName string) {
+	_, isBuiltin := builtinNames()[agentName]
+	src := func() string {
+		if isBuiltin {
+			return "built-in"
+		}
+		return "global"
+	}
+
+	override := resolved.merged.Agents[agentName]
+
+	_, _ = fmt.Fprintf(w, "Name:        %s\n", ag.Name)
+	_, _ = fmt.Fprintf(w, "Image:       %s  [%s]\n", ag.Image, fieldSourceImage(agentName, resolved, isBuiltin))
+	_, _ = fmt.Fprintf(w, "Command:     %s  [%s]\n", strings.Join(ag.Command, " "), src())
+	if override.Description != "" {
+		_, _ = fmt.Fprintf(w, "Description: %s\n", override.Description)
+	}
+
+	// Env: names/patterns only.
+	_, _ = fmt.Fprintf(w, "Env forward: %s\n", joinOrNone(ag.EnvForward))
+	_, _ = fmt.Fprintf(w, "Default env: %s\n", joinOrNone(sortedKeys(ag.DefaultEnv)))
+	if len(override.EnvRequired) > 0 {
+		_, _ = fmt.Fprintf(w, "Required env:\n")
+		for _, name := range override.EnvRequired {
+			_, present := os.LookupEnv(name)
+			status := "MISSING"
+			if present {
+				status = "present"
+			}
+			_, _ = fmt.Fprintf(w, "  %-30s %s\n", name, status)
+		}
+	}
+
+	// Resources.
+	_, _ = fmt.Fprintf(w, "Resources:   %d CPUs, %s memory, %s /tmp\n",
+		ag.DefaultCPUs, ag.DefaultMemory, ag.DefaultTmpSize)
+
+	// Egress.
+	_, _ = fmt.Fprintf(w, "Egress:      profile=%s\n", ag.DefaultEgressProfile)
+	for _, profile := range sortedProfiles(ag.EgressHosts) {
+		names := make([]string, 0, len(ag.EgressHosts[profile]))
+		for _, h := range ag.EgressHosts[profile] {
+			names = append(names, h.Name)
+		}
+		_, _ = fmt.Fprintf(w, "  %-12s %s\n", string(profile)+":", strings.Join(names, ", "))
+	}
+
+	// MCP.
+	mode := ""
+	authz := ""
+	enabled := "inherit"
+	mcpDisabled := false
+	if override.MCP != nil {
+		mode = override.MCP.Mode
+		if override.MCP.Enabled != nil {
+			if *override.MCP.Enabled {
+				enabled = "true"
+			} else {
+				enabled = "false"
+				mcpDisabled = true
+			}
+		}
+		if override.MCP.Authz != nil {
+			authz = override.MCP.Authz.Profile
+		}
+	}
+	// A custom agent with MCP enabled and no explicit authz runs under the
+	// safe-tools default applied at the composition root (built-ins keep
+	// full-access). Surface that effective default here so inspect reflects
+	// the profile the agent will actually run with.
+	if authz == "" && !isBuiltin && !mcpDisabled {
+		authz = domainconfig.DefaultCustomAgentMCPAuthzProfile + " (default)"
+	}
+	_, _ = fmt.Fprintf(w, "MCP:         enabled=%s mode=%q authz=%q\n", enabled, mode, authz)
+
+	// Credential paths.
+	_, _ = fmt.Fprintf(w, "Credentials: %s\n", joinOrNone(ag.CredentialPaths))
+
+	// Settings mappings.
+	if ag.SettingsManifest != nil && len(ag.SettingsManifest.Entries) > 0 {
+		_, _ = fmt.Fprintf(w, "Settings:\n")
+		for _, e := range ag.SettingsManifest.Entries {
+			_, _ = fmt.Fprintf(w, "  [%s] %s -> %s (%s%s%s)\n",
+				e.Category, e.HostPath, e.GuestPath, kindString(e.Kind),
+				formatSuffix(e.Format), optionalSuffix(e.Optional))
+		}
+	} else {
+		_, _ = fmt.Fprintf(w, "Settings:    (none)\n")
+	}
+}
+
+func agentsDoctorCmd() *cobra.Command {
+	var cfgPath string
+	cmd := &cobra.Command{
+		Use:   "doctor <name>",
+		Short: "Validate an agent's configuration",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runAgentsDoctor(cmd, cfgPath, args[0])
+		},
+	}
+	cmd.Flags().StringVar(&cfgPath, "config", "", "Config file path (default: ~/.config/broodbox/config.yaml)")
+	return cmd
+}
+
+func runAgentsDoctor(cmd *cobra.Command, cfgPath, agentName string) error {
+	ws, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("getting current directory: %w", err)
+	}
+	resolved, err := buildResolvedRegistry(cfgPath, ws, slog.Default(), io.Discard)
+	if err != nil {
+		return err
+	}
+
+	// A misconfigured custom agent is SKIPPED during registration (registerCustomAgents
+	// warns and drops it), so it is absent from the registry. Diagnosing exactly
+	// that case is the whole point of doctor — so do not hard-gate on the registry.
+	// Only reject names that are neither registered nor declared in the merged config.
+	_, regErr := resolved.registry.Get(agentName)
+	_, declared := resolved.merged.Agents[agentName]
+	if regErr != nil && !declared {
+		return fmt.Errorf("unknown agent %q", agentName)
+	}
+
+	override := resolved.merged.Agents[agentName]
+	localAddedAgent := didLocalAddAgent(resolved, agentName)
+	localAddedCreds := didLocalAddCredentials(resolved, agentName)
+	_, isBuiltin := builtinNames()[agentName]
+
+	results, ok := runDoctor(agentName, override, isBuiltin, os.LookupEnv, imageRefValidator, localAddedAgent, localAddedCreds)
+
+	out := cmd.OutOrStdout()
+	for _, r := range results {
+		status := "PASS"
+		if !r.ok {
+			status = "FAIL"
+		}
+		_, _ = fmt.Fprintf(out, "[%s] %s\n", status, r.message)
+	}
+	if !ok {
+		return errors.New("agent configuration has problems (see above)")
+	}
+	return nil
+}
+
+// checkResult is a single doctor check outcome.
+type checkResult struct {
+	ok      bool
+	message string
+}
+
+// runDoctor performs all agent health checks against the resolved override. It
+// is pure (the environment and image-ref lookups are injected) so it can be
+// exercised directly in tests without touching the real config tree. It
+// returns the per-check results and an overall ok flag (false if any FAIL).
+//
+// The custom-agent identity checks (name/image/command/paths/egress via
+// ValidateCustomAgent) only apply to custom (bring-your-own) agents declared
+// in config. Built-in agents carry their image/command in the registry entry,
+// not in the override, so feeding them through ValidateCustomAgent would
+// spuriously fail on the empty Image. For built-ins we skip those checks and
+// only run the override-applicable checks (required env, blocked local
+// additions).
+//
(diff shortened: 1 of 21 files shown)


pkg/domain/config/custom_agent.go
```diff
@@ -0,0 +1,314 @@
...
+
+	return nil
+}
+
+// safeRelPath validates that p is a non-empty relative path that does not
+// escape the sandbox home directory: no leading "/", no ".." element, and not
+// "." after cleaning. This is the single source of truth for path safety on
+// custom-agent credential and settings paths.
+func safeRelPath(p string) error {
+	if p == "" {
```
JAORMX, line 292: **[High — Spec] The spec's canonical Phase 1 `aider` example does not boot**

The issue's acceptance-criteria example declares only `image`, `command`, `env_forward`, and `mcp.mode: env` — no `egress_profile`, no `egress_hosts`:

```yaml
agents:
  aider:
    image: ghcr.io/acme/aider-bbox:latest
    command: ["aider"]
    env_forward: [OPENAI_API_KEY, "AIDER_*"]
    mcp:
      mode: env
```

`ValidateCustomAgent` defaults the effective egress profile to `standard` (non-permissive) and then **rejects** any non-permissive profile that has no `egress_hosts` entry for that profile:

```go
if effectiveProfile != egress.ProfilePermissive {
    if len(o.EgressHosts[string(effectiveProfile)]) == 0 {
        return fmt.Errorf("agent %q: egress profile %q requires egress_hosts[%q] (or set egress_profile: permissive)", ...)
    }
}
```

So the exact config the issue says should work is rejected by `bbox agents doctor aider` and `registerCustomAgents` drops the agent at startup. The PR body's own `bbox aider` smoke test must have used a richer config than the minimal acceptance example.

This is a spec-vs-implementation mismatch on Phase 1's headline acceptance criterion. Two reasonable resolutions:
1. Loosen `ValidateCustomAgent`: don't require `egress_hosts` for `standard` when the agent also declares `mcp.mode: env` (the MCP proxy is the only network path that matters), OR
2. Update the issue's acceptance example to include `egress_hosts`, and document that `standard` always requires hosts (the safer default, but it changes the spec's "just works" promise).

pkg/domain/config/config.go
```diff
@@ -1059,6 +1194,9 @@ func mergeAgentOverride(global, local AgentOverride) AgentOverride {
 		if local.MCP.Enabled != nil {
 			result.MCP.Enabled = local.MCP.Enabled
 		}
+		if local.MCP.Mode != "" {
+			result.MCP.Mode = local.MCP.Mode
```
JAORMX: **[Medium — secure-code-reviewer] CWE-862 / OWASP A01:2021 — Workspace-local `.broodbox.yaml` can override a per-agent `mcp.mode`, suppressing a built-in's config-file injector**

`mergeAgentOverride` copies `local.MCP.Mode` verbatim:
```go
if local.MCP.Mode != "" {
    result.MCP.Mode = local.MCP.Mode
}
```
And `sandbox.go` then drops the built-in injector when mode is `env`:
```go
if mcpMode(cfg, agentName) == config.MCPModeEnv {
    mcpInjector = nil
}
```

**Trigger:** A workspace `.broodbox.yaml` shipped by an untrusted repo sets `agents.claude-code.mcp.mode: "env"`. After merge, the built-in plugin's `MCPConfig()` injector is discarded and claude-code only learns the proxy via `BBOX_MCP_URL`. For agents that don't read that env var (or read it incorrectly), the proxy effectively disappears — a **fail-open** outcome from untrusted input.

This is the **only** per-agent MCP field where workspace-local config changes built-in agent *behavior* rather than tightening it. Every other per-agent field is either tighten-only (`Authz`, `EgressProfile`), ignored from local (`Image`, `Command`, `Credentials`, `Settings`, `EgressHosts`), or additive-but-bounded (`AllowHosts`). The `mergeAgentOverride` doc comment advertises "local may only TIGHTEN" — this branch violates that invariant.

**Fix:** Treat `MCP.Mode` from workspace-local config the same way `Image`/`Command` are treated — ignore it from local, source only from global config. Drop the `local.MCP.Mode` branch entirely.

**Verification:** Regression test where `local.Agents["claude-code"].MCP.Mode = "env"` leaves `merged.Agents["claude-code"].MCP.Mode == ""` (mirroring the existing `local image override is ignored (security #7)` test in `config_test.go:131`).

pkg/domain/config/config.go
```diff
@@ -1059,6 +1194,9 @@ func mergeAgentOverride(global, local AgentOverride) AgentOverride {
 		if local.MCP.Enabled != nil {
```
JAORMX: **[Low — secure-code-reviewer] CWE-862 / OWASP A05:2021 — Workspace-local `.broodbox.yaml` can re-enable a per-agent MCP proxy the operator disabled globally**

`mergeAgentOverride` copies `local.MCP.Enabled` verbatim:
```go
if local.MCP.Enabled != nil {
    result.MCP.Enabled = local.MCP.Enabled
}
```
And `resolveMCPConfig` lets per-agent `Enabled` override the global gate:
```go
if override.MCP != nil && override.MCP.Enabled != nil {
    mcpCfg.Enabled = override.MCP.Enabled
}
```

**Trigger:** Operator sets `mcp.enabled: false` globally for an agent. Workspace `.broodbox.yaml` sets `agents.<name>.mcp.enabled: true` → proxy starts for that agent even though the operator turned MCP off.

This breaks the "local can only tighten, never widen" invariant. The CLI `--no-mcp` path wins separately (checked in `main.go:785-789`), so the residual exposure is narrower: an operator who disables MCP only via global *config* (not the flag) can be re-enabled by an untrusted repo. The MCP authz profile stays tighten-only, so this is "proxy turned back on," not privilege escalation — but it's inconsistent with how `SettingsImport.Enabled` is handled (strictly tighten-only at `mergeSettingsImportConfig`).

**Fix:** Apply tighten-only semantics to `MCP.Enabled` per-agent: workspace-local can only disable (`Enabled: false`), never re-enable. Mirror `mergeSettingsImportConfig`. Regression test: `global.MCP.Enabled = false`, `local.MCP.Enabled = true` → merged stays `false`.

cmd/bbox/main.go
```diff
@@ -1131,6 +1085,138 @@ func chooseObserver(terminal *infraterminal.OSTerminal) progress.Observer {
...
+			// Already a built-in agent — overrides are applied at run time.
+			continue
+		}
+		if override.Image == "" {
+			// Not a custom-agent definition (e.g. an override for an agent
+			// that is not currently registered). Skip silently.
+			continue
+		}
+		if err := domainconfig.ValidateCustomAgent(name, override, imageRefValidator); err != nil {
+			_, _ = fmt.Fprintf(warnWriter, "Warning: skipping custom agent %q: %s\n", name, err)
```
JAORMX, line 1241: **[Medium — ux-reviewer] `bbox <declared-but-skipped-agent>` is a dead-end that names the wrong cause**

When a custom agent fails `ValidateCustomAgent`, `registerCustomAgents` prints a `Warning: skipping custom agent %q: …` line to stderr (`main.go:1206`) and drops it from the registry. Then `bbox my-agent` hits `registry.Get` → `ErrNotFound`, and the error shown is:

```
Error: resolving agent: agent not found: "my-agent"

Available agents:
  claude-code  …
  codex        …
```

From the user's seat this reads as "you typo'd the name" — but they *did* declare it. The earlier warning scrolls away, the "Available agents" list pointedly omits their agent, and nothing connects the two facts or points at `bbox agents doctor my-agent` (the command built precisely for this). The merged config still carries the key (`registerCustomAgents` never removes it), so `run()` can detect "declared in `cfg.Agents` but not in registry" and route the error to doctor.

**Fix:** In the `registry.Get` not-found path, check whether the name exists as a key in `cfg.Agents`; if so, emit a targeted error pointing at `bbox agents doctor <name>` (and surface the validation failure inline if it's still in scope).

cmd/bbox/agents.go
```diff
@@ -0,0 +1,466 @@
...
+		add(true, "built-in agent: identity provided by the registry (no custom config to validate)")
+	} else if err := domainconfig.ValidateCustomAgent(agentName, override, imageValidator); err != nil {
+		// The bulk of validation is the same pure check the loader runs.
+		add(false, fmt.Sprintf("config validation: %s", err))
+	} else {
+		add(true, "config validation: name, image, command, paths, egress all valid")
+	}
+
+	// Required env presence (names only — values never read into output).
+	for _, name := range override.EnvRequired {
```
JAORMX, line 320: **[Medium — ux-reviewer] `env_required` is not enforced at run time — only by a separate command**

`EnvRequired` is checked by `agents inspect`/`doctor` (`agents.go:141`, `agents.go:315`) but **nowhere** on the `bbox <agent>` run path (`pkg/sandbox/sandbox.go` `Prepare` never references it). So a user who configures `env_required: [ACME_API_KEY]`, forgets to export it, and runs `bbox my-agent` gets a full VM boot followed by an opaque in-VM failure — the very footgun `env_required` exists to prevent.

**Fix:** Gate required-env presence in `Prepare` (fail fast before booting the VM, pointing to `bbox agents doctor <name>`). This is a one-line loop over `override.EnvRequired` calling `os.LookupEnv`, returning an actionable error if any is absent. Stance #4 (actionable errors) applies.

cmd/bbox/agents.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	// Env: names/patterns only.
+	_, _ = fmt.Fprintf(w, "Env forward: %s\n", joinOrNone(ag.EnvForward))
+	_, _ = fmt.Fprintf(w, "Default env: %s\n", joinOrNone(sortedKeys(ag.DefaultEnv)))
+	if len(override.EnvRequired) > 0 {
+		_, _ = fmt.Fprintf(w, "Required env:\n")
+		for _, name := range override.EnvRequired {
+			_, present := os.LookupEnv(name)
+			status := "MISSING"
```
JAORMX: **[Medium — ux-reviewer] MISSING marker gives no next action**

```
Required env:
  ACME_API_KEY                    MISSING
```

"MISSING" names the problem but not the fix. A one-line suffix (`— set it on the host, e.g. export ACME_API_KEY=…`) or a trailing `Run 'bbox agents doctor acme' for a full check` would close the loop. Same gap in `agents doctor`: `required env ACME_API_KEY is missing` states the fact with no remedy.

cmd/bbox/agents.go
```diff
@@ -0,0 +1,466 @@
...
+		return fmt.Errorf("getting current directory: %w", err)
+	}
+	resolved, err := buildResolvedRegistry(cfgPath, ws, slog.Default(), io.Discard)
+	if err != nil {
+		return err
+	}
+	out := cmd.OutOrStdout()
+	for _, e := range resolved.registry.List() {
+		_, _ = fmt.Fprintf(out, "%-15s %s\n", e.Agent.Name, e.Agent.Image)
+	}
```
JAORMX, line 69: **[Low — ux-reviewer] `agents list` doesn't distinguish built-in from custom**

```go
for _, e := range resolved.registry.List() {
    _, _ = fmt.Fprintf(out, "%-15s %s\n", e.Agent.Name, e.Agent.Image)
}
```

The PR's own `inspect`/`doctor` logic carefully separates built-in/global/workspace provenance, but `list` — the first command a user reaches for — flattens them all into one undifferentiated column. A user with no custom agents configured sees five rows and cannot tell whether *any* custom agents exist, let alone which.

**Fix:** Add a `TYPE` column (built-in / custom) or a section break. Trivial change, high orienting value.

cmd/bbox/agents.go
```diff
@@ -0,0 +1,466 @@
...
+	src := func() string {
+		if isBuiltin {
+			return "built-in"
+		}
+		return "global"
+	}
+
+	override := resolved.merged.Agents[agentName]
+
+	_, _ = fmt.Fprintf(w, "Name:        %s\n", ag.Name)
```
JAORMX, line 136: **[Low — ux-reviewer] Provenance labels exist only for Image and Command**

`renderInspect` brackets `[built-in]`/`[global]`/`[workspace]` on Image and Command only. EnvForward, DefaultEnv, Resources, Egress profile/hosts, MCP, Credentials, and Settings carry no source label — yet these are the fields a user debugging "where did this egress host / env forward come from" most needs to trace. The PR claim of "field provenance" is met for two fields and unmet for the rest.

At minimum, egress profile and env_forward provenance would help, since workspace config can tighten those.

cmd/bbox/agents.go
```diff
@@ -0,0 +1,466 @@
...
+
+	entry, err := resolved.registry.Get(agentName)
+	if err != nil {
+		// Not registered. If it is a custom agent declared in the merged config
+		// (skipped during registration because it failed validation), build a
+		// best-effort view from the override so the user can still see what they
+		// declared. Direct them to `agents doctor` for the failure detail.
+		override, declared := resolved.merged.Agents[agentName]
+		if !declared {
+			return fmt.Errorf("unknown agent %q", agentName)
```
JAORMX: **[Low — ux-reviewer] "unknown agent %q" in inspect/doctor points nowhere**

```go
return fmt.Errorf("unknown agent %q", agentName)
```

A trailing `— run 'bbox agents list' to see available agents` turns a dead-end into a path. (Same at `agents.go:242` for doctor.)

JAORMX (commented):
## Panel review — 3 axes, 5 reviewers

Ran a three-axis panel review (Spec / Standards / Domain) with the Domain axis fanned out to `secure-code-reviewer`, `software-architect`, `devex-reviewer`, and `ux-reviewer`. Findings are inline as line-anchored comments; summary below. (`software-architect` and `devex-reviewer` reached their turn limits mid-investigation; their partial findings are folded into the summary where confirmed, and the DDD-layer purity check — the most important invariant — was completed by the Standards axis.)

### Spec — ⚠️ one Wrong finding against the acceptance criteria
- **[High]** The issue's canonical Phase 1 `aider` example (only `image`/`command`/`env_forward`/`mcp.mode: env`, no egress) is **rejected** by `ValidateCustomAgent` — `standard` egress profile requires `egress_hosts`, so the exact config the issue says should work doesn't boot. Needs either a looser validation or an updated acceptance example. `custom_agent.go:283`.
- **Scope creep (noted, acceptable):** dropping local-only agent keys and ignoring local `image`/`command` during merge silently changes pre-existing workspace override semantics for *all* agents, not just custom ones. Justified as security #7, but it's a contract change.

### Standards — ✅ clean (one soft note)
DDD layer purity **verified** — the most important invariant:
- `pkg/domain/config/custom_agent.go` imports only stdlib + sibling domain packages. The image-ref parser is correctly injected as `imageRefValidator func(string) error` (closure supplied at `cmd/bbox/main.go` via `name.ParseReference`) — `go-containerregistry` stays out of domain. ✓
- `pkg/domain/agent/env_bbox.go` is pure domain, no I/O. ✓
- `pkg/sandbox/sandbox.go` new imports are only domain packages. ✓
- `pkg/clients/*/mcp.go` new imports are `pkg/domain/config` (a domain package) — permitted; no `internal/infra` import. ✓

SPDX headers present on all new `.go` files. `log/slog` exclusively. Errors wrapped with `%w`. Table-driven tests alongside code. No business logic added to `internal/infra/`. One soft note: commit subject is 58 chars > 50 (CLAUDE.md convention) — judgement call, not a blocker.

### Domain — 1 High, 2 Medium, 6 Low/Info

**Ship-blockers (1)**
- **[High]** (Spec, above) Canonical acceptance example rejected by validation.

**Mechanical fixes (2)**
- **[Medium — secure-code-reviewer] CWE-862** Workspace-local `.broodbox.yaml` can override a per-agent `mcp.mode` to `"env"`, suppressing a built-in's config-file injector — fail-open from untrusted input. The only per-agent MCP field where local changes built-in behavior rather than tightening. Fix: ignore `local.MCP.Mode` (treat like `Image`/`Command`). `config.go:1198`.
- **[Medium — ux-reviewer]** `env_required` is not enforced at run time — only by `agents doctor`. A user who forgets to export a required env var gets a full VM boot then an opaque in-VM failure. Fix: gate in `Prepare`, fail fast before boot. `agents.go:315`.

**Judgement calls (3)**
- **[Low — secure-code-reviewer] CWE-862/A05** Workspace-local can re-enable a per-agent MCP proxy the operator disabled globally (via config, not `--no-mcp`). Breaks the "local can only tighten" invariant; inconsistent with `SettingsImport.Enabled`. Fix: tighten-only on `MCP.Enabled`. `config.go:1194`.
- **[Medium — ux-reviewer]** `bbox <declared-but-skipped-agent>` dead-ends on "agent not found" with no mention of the earlier validation warning or `bbox agents doctor`. The merged config still carries the key, so `run()` can detect "declared but not registered" and route to doctor. `main.go:1206`.
- **[Low — ux-reviewer]** `agents inspect` MISSING marker names the problem but not the fix. `agents.go:145`.

**Polish (3)**
- **[Low — ux-reviewer]** `agents list` doesn't distinguish built-in from custom — add a `TYPE` column or section break. `agents.go:64`.
- **[Low — ux-reviewer]** Provenance labels exist only for Image and Command — extend to egress profile / env_forward at minimum. `agents.go:131`.
- **[Low — ux-reviewer]** "unknown agent %q" in inspect/doctor points nowhere — append `run 'bbox agents list'`. `agents.go:100`.

### Verified security claims (secure-code-reviewer, no action required)
For the record, since the PR body makes explicit security claims I verified each:
1. ✓ Tighten-only merge (env_forward subset-coverage with glob patterns is sound — `AIDER_SECRET_*` ⊆ `AIDER_*` covered, reverse rejected)
2. ✓ Path validation rejects `../../etc/shadow`, `/etc/shadow` (defense-in-depth in `FSStore`/`FSInjector` too)
3. ✓ `env_forward` rejects empty/bare-`*`/leading-star at every load site (global, workspace, CLI, SDK)
4. ✓ `BBOX_*` env applied after forwarded host vars (authoritative; regression test at `sandbox_test.go:1182`)
5. ✓ MCP authz profile tighten-only (workspace cannot widen to `full-access`)
6. ✓ `agents inspect` never leaks env values (names/patterns + present/MISSING only; regression test at `agents_test.go:140`)
7. ✓ Image-ref parser injection has no injection vector (syntactic validation only, no network I/O, not constructible from config)

### Gaps
- `software-architect` and `devex-reviewer` hit turn limits mid-investigation. The DDD-layer purity check (their primary focus) was completed by the Standards axis and is clean. Any deeper architectural concerns about the `SandboxConfig` contract changes or `BuildUniversalEnv` signature ergonomics would need a follow-up pass.

**Most important single issue:** the High spec mismatch — the acceptance example doesn't boot. The Medium `mcp.mode` override (secure-code-reviewer) is the highest-confidence security finding since it's the only field that breaks the PR's own advertised "local may only tighten" invariant.

pkg/domain/config/custom_agent.go
JAORMX: Addressed in the review-fixes commit (now `015b6c3` after rebase onto `main`).

`ValidateCustomAgent` no longer requires `egress_hosts` for a non-permissive profile when the agent also declares `mcp.mode: env` — the MCP proxy is the agent's network discovery path, so a hostless `standard` config is legitimate at load time (`pkg/domain/config/custom_agent.go:275-282`).

The runtime safety net still holds: `egress.Resolve` rejects a hostless non-permissive profile at VM start, so the load-time loosening alone can't boot the canonical example. `sandbox.go` `Prepare` was updated to fall back to an empty (gateway-only) restricted policy in exactly that case — all external egress stays blocked, the proxy is the only path out (`pkg/sandbox/sandbox.go:387-392`). A real runtime boot test (`TestSandboxRunner_Prepare_MCPModeEnvHostlessStandardBoots`) asserts the policy is non-nil but empty.

The issue's canonical `aider` example now boots end-to-end.

pkg/domain/config/config.go
JAORMX: Addressed in the review-fixes commit (now `015b6c3`).

`mergeAgentOverride` now ignores `local.MCP.Mode` entirely — it is sourced from global config only, exactly like `Image`/`Command`. The `local.MCP.Mode` copy branch was dropped (`pkg/domain/config/config.go:1191-1213`). A workspace `.broodbox.yaml` can no longer flip a built-in agent's MCP mode to `env` and suppress its config-file injector.

Regression test added: `TestMergeAgentOverride/ignores_local_mcp_mode_(security)` asserts `local.MCP.Mode = "env"` leaves `merged.MCP.Mode == ""`.

pkg/domain/config/config.go
JAORMX: Addressed in the review-fixes commit (now `015b6c3`).

`MCP.Enabled` is now tighten-only per-agent, mirroring `SettingsImport.Enabled`: workspace-local may only DISABLE (`Enabled: false`), never re-enable what the operator disabled globally (`pkg/domain/config/config.go:1209-1210`).

Regression test added: `TestMergeAgentOverride/mcp.enabled_tighten-only_(security)` asserts `global.MCP.Enabled=false`, `local.MCP.Enabled=true` → merged stays `false`.

cmd/bbox/main.go
JAORMX: Addressed in the review-fixes commit (now `015b6c3`).

The `registry.Get` not-found path in `run()` now checks whether the name exists as a key in the merged `cfg.Agents`. If it does (declared but skipped during registration because it failed validation), the error points at `bbox agents doctor <name>` instead of the generic "you typo'd the name" `Available agents` list (`cmd/bbox/main.go:1043-1059`).

cmd/bbox/agents.go
JAORMX: Addressed in the review-fixes commit (now `015b6c3`).

`env_required` is now enforced in `Prepare` before booting the VM: a missing required env var fails fast with a pointer to `bbox agents doctor <name>` (`pkg/sandbox/sandbox.go:321-337`). Uses the injected `EnvProvider` (no direct `os.Getenv`) to stay testable and layer-clean.

Test: `TestSandboxRunner_Prequire_MissingRequiredEnv` asserts the fail-fast path returns the actionable error before any VM starts.

cmd/bbox/agents.go
JAORMX: Addressed in the review-fixes commit (now `015b6c3`).

`agents list` now prints a `TYPE` column distinguishing `built-in` from `custom` (`cmd/bbox/agents.go:63-68`).

cmd/bbox/agents.go
JAORMX: Addressed in the review-fixes commit (now `015b6c3`).

Provenance labels extended beyond Image/Command:
- `EnvForward` — `fieldSourceEnvForward` (`cmd/bbox/agents.go:144, 381-403`)
- `EgressProfile` — `fieldSourceEgressProfile` (`cmd/bbox/agents.go:163, 405-427`)

Both honor the tighten-only merge: a workspace-local override that narrows/tightens shows as `workspace`; otherwise `global` (or `built-in` for registry agents with no override).

PR #207, opened 2026-06-30
Add bbox run-image ephemeral one-shot command

### LINKED ISSUE #205: BYO agents: `bbox run-image` ephemeral one-shot run
Follow-up to #191 (epic), UX Tier 1. Lets a user experiment with any image once, without editing config or registering an agent.

## Scope

```bash
bbox run-image ghcr.io/me/aider-bbox:latest -- aider --model gpt-5

bbox run-image ghcr.io/me/aider-bbox:latest \
  --name aider --env OPENAI_API_KEY --env 'AIDER_*' \
  --memory 4g --cpus 4 --mcp --mcp-authz-profile safe-tools \
  --egress-profile standard --allow-host api.openai.com:443 -- aider
```

This mode is **explicitly ephemeral**:
- no credential persistence by default
- no settings import by default
- no custom config-file injection unless requested
- no env forwarding unless explicitly requested

## Notes / constraints

- Builds an in-memory `agent.Agent` from flags via the existing `config.AgentFromOverride` path (#200) and runs it through the normal `SandboxRunner` — no config-file mutation, no registry persistence.
- Snapshot isolation, review, egress/MCP tighten-only semantics all still apply.
- Universal `BBOX_*` env vars (#200) are injected as for any agent.
- Define the **minimum image contract** for an arbitrary image (boot/init compatibility, sandbox user/home, shell, command on PATH) — coordinate with the OCI-manifest follow-up (Phase 5).

## Acceptance criteria

`bbox run-image IMAGE -- CMD` boots the image, runs the command, forwards only explicitly requested env, and persists nothing.

JAORMX:
## Summary

Implements #205 — `bbox run-image IMAGE [flags] -- CMD`, an ephemeral one-shot that boots an arbitrary OCI image in a sandbox VM, runs the command, forwards only explicitly-requested env, and persists nothing. No config-file mutation, no registry persistence.

```bash
bbox run-image ghcr.io/me/aider-bbox:latest -- aider --model gpt-5

bbox run-image ghcr.io/me/aider-bbox:latest \
  --name aider --env OPENAI_API_KEY --env 'AIDER_*' \
  --memory 4g --cpus 4 --mcp --mcp-authz-profile safe-tools \
  --egress-profile standard --allow-host api.openai.com:443 -- aider
```

Builds an in-memory `agent.Agent` from flags via the existing `config.AgentFromOverride` path (#200) and runs it through the normal `SandboxRunner`.

## Ephemeral defaults (safer-than-custom)
| Feature | Default | Opt-in |
|---|---|---|
| Credential persistence | OFF | (ephemeral by design) |
| Settings import | OFF | (ephemeral by design) |
| Env forwarding | EMPTY | `--env` (repeatable; `--env-forward` alias) |
| MCP proxy | OFF | `--mcp` (forces `mcp.mode=env`; inherits `safe-tools` authz default) |
| Git token forwarding | OFF | `--git-token` |
| SSH agent forwarding | OFF | `--git-ssh-agent` |
| Egress | `permissive` (with stderr disclosure) | `--egress-profile standard --allow-host ...` |

Snapshot isolation, review, egress/MCP tighten-only semantics all still apply. Universal `BBOX_*` env vars are injected as for any agent.

## How it works
- New cobra subcommand `run-image` (cmd/bbox/run_image.go). The first positional is the image; everything after `--` is the command.
- Reuses `run()` by extracting a shared `runSandbox(...)` tail (cmd/bbox/main.go); `run()` keeps its front-end, `runRunImage()` builds the in-memory agent then calls the same tail.
- The ephemeral agent is registered data-only (nil Plugin) via `registry.Add` — same seam `registerCustomAgents` uses — so `isCustomAgent`/`applyCustomAgentAuthzDefault` apply the `safe-tools` MCP authz default. Nothing is written to disk.
- Name derives from the image repo basename (sanitized) when `--name` is unset; validated via `agent.ValidateName`.
- Pure helpers `buildRunImageOverride` and `deriveAgentName` are table-tested; `validateMCPFlags` extracted to de-duplicate front-end validation.

## Minimum image contract
Documented in `docs/run-image.md`: the image provides the Linux rootfs; the agent command must be on `PATH` inside the image; `/bin/sh` and a CA bundle at `/etc/ssl/certs` are expected; sandbox user UID 1000 / `/home/sandbox`. `bbox-init` is embedded by the VM runtime (PID 1), not the image.

## Verification
- `task fmt` — clean
- `task lint` — 0 issues
- `task test` (race) — all green, incl. new `cmd/bbox` tests
- New tests: `deriveAgentName`, `buildRunImageOverride` (defaults/env/mcp/authz/egress/resources/copy-semantics), `ValidateCustomAgent` accept/reject, cobra arg parsing (`IMG -- CMD`, missing `--`, empty command), RunE pre-IO errors (`--seed-credentials` rejection, invalid image ref with lowercase hint, invalid `--name`), name-collision guard (built-in + custom + derived), `--name` wins-over-derivation, `--env`/`--env-forward` alias merge, `validateMCPFlags`, and the `cfg.Agents` map-detach (no aliasing into global config).

## Review-panel driven changes
A spec + standards + domain (security/arch/UX/DevEx/duplication/QA) panel reviewed the diff; this commit applies the cross-confirmed findings: git-token/SSH-agent off-by-default (contradicts the issue's "forwards only explicitly requested env"), `--seed-credentials` rejected as incompatible, name-collision guard covers custom agents (not just built-ins), dead `name` param dropped, derived agent name surfaced on stderr, MCP validation de-duplicated, `cfg.Agents` map-aliasing fixed, image-contract discoverable from `--help`, image-ref error hints lowercase refs.

## Notes / out of scope
- `runSandbox` still hard-codes `infravm.NewMicroVMRunner` (no VMRunner injection). A full Prepare-boundary E2E test isn't reachable without that refactor; RunE-level tests stop at the pre-`runSandbox` boundary (guard, flag parsing, pure helpers). Left for a follow-up.
- The OCI-manifest minimum-image-contract verification (Phase 5 follow-up) is documented but not enforced here.

Closes #205.

### DIFF
--- a/cmd/bbox/run_image.go
+++ b/cmd/bbox/run_image.go
@@ -1,0 +1,672 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"context"
+	"crypto/rand"
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/spf13/cobra"
+	"go.opentelemetry.io/otel"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+
+	infraagent "github.com/stacklok/brood-box/internal/infra/agent"
+	"github.com/stacklok/brood-box/internal/infra/exclude"
+	infraprogress "github.com/stacklok/brood-box/internal/infra/progress"
+	infraterminal "github.com/stacklok/brood-box/internal/infra/terminal"
+	infratracing "github.com/stacklok/brood-box/internal/infra/tracing"
+	infravm "github.com/stacklok/brood-box/internal/infra/vm"
+	infraws "github.com/stacklok/brood-box/internal/infra/workspace"
+	"github.com/stacklok/brood-box/pkg/domain/agent"
+	"github.com/stacklok/brood-box/pkg/domain/bytesize"
+	domainconfig "github.com/stacklok/brood-box/pkg/domain/config"
+	"github.com/stacklok/brood-box/pkg/domain/egress"
+	"github.com/stacklok/brood-box/pkg/domain/snapshot"
+	"github.com/stacklok/brood-box/pkg/sandbox"
+)
+
+// runImageFlags holds the flag values for `bbox run-image`. It is a subset of
+// runFlags tailored to the ephemeral one-shot UX: MCP defaults to OFF (opt-in
+// via --mcp), env forwarding defaults to EMPTY, and credential/settings
+// persistence is OFF. The fields map directly onto runFlags when the shared
+// runSandbox tail is invoked.
+type runImageFlags struct {
+	name            string
+	envForward      []string
+	memory          string
+	cpus            uint32
+	tmpSize         string
+	mcp             bool
+	mcpAuthzProfile string
+	mcpGroup        string
+	mcpPort         uint16
+	mcpConfig       string
+	mcpSessionTTL   time.Duration
+	egressProfile   string
+	allowHosts      []string
+	workspace       string
+	workspaceMode   string
+	review          bool
+	excludes        []string
+	yesAckDirect    bool
+	sshPort         uint16
+	ports           []string
+	cfgPath         string
+	debug           bool
+	logFile         string
+	noFirmwareDL    bool
+	noImageCache    bool
+	pull            string
+	traceEnabled    bool
+	timings         bool
+	// gitToken / gitSSHAgent are opt-in (default false): the inverse of the
+	// root command's --no-git-token / --no-git-ssh-agent. run-image forwards
+	// neither the git token nor the SSH agent into an arbitrary image unless
+	// explicitly requested (ephemeral, safer-by-default posture).
+	gitToken        bool
+	gitSSHAgent     bool
+	seedCredentials bool
+}
+
+// runImageCmd builds the `bbox run-image` subcommand. It boots an arbitrary
+// OCI image in a sandbox VM, runs the supplied command, forwards only
+// explicitly-requested env, and persists nothing — an ephemeral one-shot with
+// safer-than-custom defaults. The image is the first positional arg; the
+// command (required) follows a literal `--`.
+func runImageCmd() *cobra.Command {
+	var (
+		f         runImageFlags
+		envFwd    []string
+		envFwdAlt []string
+	)
+	cmd := &cobra.Command{
+		Use:   "run-image IMAGE [flags] -- CMD [args...]",
+		Short: "Run an arbitrary OCI image as an ephemeral sandbox one-shot",
+		Long: `bbox run-image boots an arbitrary OCI image in a hardware-isolated
+microVM, runs CMD inside it, and persists nothing. It builds an in-memory
+agent from the flags (no config-file mutation, no registry persistence) and
+runs it through the normal sandbox path.
+
+Ephemeral defaults (safer than a custom agent):
+  - credential persistence OFF
+  - host settings import OFF
+  - env forwarding EMPTY (forward nothing) — opt in with --env
+  - git token / SSH agent OFF — opt in with --git-token / --git-ssh-agent
+  - MCP proxy OFF — opt in with --mcp
+  - egress PERMISSIVE (disclosed on stderr) — tighten with --egress-profile
+
+The image is the first positional argument. The command to run inside the VM
+follows a literal "--" and is required.
+
+Example:
+  bbox run-image ghcr.io/jbarslox/aider-bbox:latest -- aider
+  bbox run-image ubuntu:24.04 --env OPENAI_API_KEY --egress-profile standard --allow-host api.openai.com:443 -- python -m http.server
+
+See: docs/run-image.md (minimum image contract)`,
+		Args: cobra.MinimumNArgs(1),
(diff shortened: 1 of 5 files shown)


cmd/bbox/run_image.go
```diff
@@ -0,0 +1,672 @@
...
+	}
+
+	// Egress profile defaults to permissive for the ephemeral UX; an explicit
+	// --egress-profile (validated upstream) wins. An empty f.egressProfile
+	// cannot happen (the flag default is "permissive"), but guard anyway.
+	profile := f.egressProfile
+	if profile == "" {
+		profile = string(egress.ProfilePermissive)
+	}
+	o.EgressProfile = profile
```
jhrozek, line 571: **HIGH — `--allow-host` never reaches the override, so `--egress-profile standard/locked` is unusable.**

`buildRunImageOverride` sets `o.EgressProfile` here but never folds `f.allowHosts` into `o.EgressHosts`. `--allow-host` flows only into `RunOpts.AllowHosts` (applied at VM start). But `ValidateCustomAgent` runs first (run_image.go:350) and rejects any non-permissive profile with no `EgressHosts[profile]` entry (custom_agent.go:276-282).

Since `run-image` has no config file to declare `egress_hosts`, `--allow-host` is the *only* way to supply them — so no non-permissive egress profile works unless `--mcp` is also passed (which trips the `mcpModeEnv` exception). This breaks the documented example in `--help`, `README.md`, and `docs/run-image.md`:

```
bbox run-image ubuntu:24.04 --env OPENAI_API_KEY --egress-profile standard --allow-host api.openai.com:443 -- python -m http.server
```

Reproduced:
```
agent "ubuntu": egress profile "standard" requires egress_hosts["standard"] (or set egress_profile: permissive)
```

**Suggested fix** — fold allow-hosts into the override under the selected profile:
```go
if len(f.allowHosts) > 0 {
    o.EgressHosts = map[string][]string{profile: append([]string(nil), f.allowHosts...)}
}
```
`--allow-host` values like `api.openai.com:443` are exactly what `ToEgressHosts` expects. The duplicate application via `RunOpts.AllowHosts` is harmless, or clear `runFlags.allowHosts` once it's in the override. Please add a regression test asserting `ValidateCustomAgent` passes for `standard + --allow-host` without `--mcp`.

cmd/bbox/run_image.go
```diff
@@ -0,0 +1,672 @@
...
+			return domainconfig.AgentOverride{}, fmt.Errorf("--tmp-size: %w", err)
+		}
+		o.TmpSize = parsed
+	}
+
+	// MCP: --mcp forces mode=env so the proxy is exposed only via BBOX_MCP_URL
+	// (no config-file injector for an arbitrary image). Authz is resolved at
+	// runSandbox wiring; the per-agent Authz override is set only when the
+	// operator supplied --mcp-authz-profile.
+	if f.mcp {
```
jhrozek, line 614: **MEDIUM — MCP sub-flags are silently ignored without `--mcp`.**

This block is gated on `if f.mcp`, so when `--mcp` is unset `o.MCP` stays nil and `--mcp-authz-profile`, `--mcp-group`, `--mcp-port`, `--mcp-config`, and `--mcp-session-ttl` are all silently dropped. The command even *validates* `--mcp-authz-profile` / `--mcp-session-ttl` up front (run_image.go:248), which implies they take effect.

This is a security-flavored footgun: `bbox run-image IMG --mcp-authz-profile observe -- cmd` runs with no MCP proxy and no warning, so the operator may believe they restricted tool access when nothing happened. (The root `run` command isn't affected — MCP is on by default there.)

**Suggested fix** — error or warn when any MCP sub-flag is set while `--mcp` is unset, consistent with the existing stderr disclosures at run_image.go:436-449:
```go
_, _ = fmt.Fprintln(os.Stderr, "Warning: --mcp-* flags ignored without --mcp")
```

jhrozek (commented):
Two findings from review (one HIGH egress bug, one MEDIUM MCP UX issue). The rest of the PR looks solid — clean `runSandbox` extraction, DDD boundaries respected, and strong table-driven tests.

cmd/bbox/run_image.go
jhrozek: Resolved in d140925.

`buildRunImageOverride` now folds each `--allow-host` value into `override.EgressHosts[profile]` (parsed via `egress.ParseHostFlag`, converted to `EgressHostConfig`), so the `ValidateCustomAgent` gate sees hosts for a non-permissive profile. `runFlags.allowHosts` is cleared to avoid filing the same hosts twice (verified `egress.Merge` in sandbox.go appends without dedup). An invalid `--allow-host` (e.g. an IP) now errors before any I/O.

Correction to my earlier snippet: the field is `map[string][]EgressHostConfig`, not `map[string][]string`, and `--allow-host` strings need `ParseHostFlag` + conversion — the committed fix does this. Regression tests cover the fold, the previously-broken `standard + --allow-host` validation (no `--mcp`), and IP rejection.

cmd/bbox/run_image.go
jhrozek: Resolved in d140925.

`RunE` now warns to stderr when MCP sub-flags (`--mcp-authz-profile`, `--mcp-group`, `--mcp-port`, `--mcp-config`, `--mcp-session-ttl`) are set without `--mcp`. Detection uses a pure helper `mcpSubFlagsWithoutMCP` backed by `cmd.Flags().Changed(...)`, so the non-zero defaults on `--mcp-group`/`--mcp-port` do not produce false positives. Covered by a hermetic table-driven test (`TestMCPSubFlagsWithoutMCP`).

Issue #201, opened 2026-06-27
BYO agents Phase 2: `bbox agents add` / `agents init` scaffolding


JAORMX:
Follow-up to #191 (epic). Phase 1 landed declarative custom agents + `bbox agents list|inspect|doctor` in #200. This issue adds the remaining Phase 2 CLI for *authoring* a custom agent without hand-editing YAML.

## Scope

- `bbox agents add NAME --image IMAGE --command CMD [--env PATTERN ...] [--memory 4g] [--cpus 4] [--egress-profile standard] [--mcp] [--mcp-authz-profile safe-tools]` — appends a validated custom agent to the **global** config file (`~/.config/broodbox/config.yaml`), then runs the existing `ValidateCustomAgent` checks and prints the result. Refuses to overwrite a built-in or an existing custom agent unless `--force`.
- `bbox agents init [NAME]` — writes a commented starter `agents:` stanza (or a standalone manifest, see #(Phase 4)) the user can fill in. Should reuse the documentation already in `internal/infra/config/writer.go`'s default template.

## Notes / constraints

- Pure mapping (`config.AgentFromOverride`) and validation (`config.ValidateCustomAgent`) already exist from #200 — reuse them; the new code is just CLI + a config-file mutation in `internal/infra/config`.
- Writing config is I/O → lives in `internal/infra/config` + `cmd/bbox`; keep the domain pure.
- Must preserve existing comments/formatting on rewrite as much as practical (or document that it normalizes).
- Custom agents remain **global-only**; `agents add` must never target a workspace `.broodbox.yaml`.

## Acceptance criteria

```bash
bbox agents add aider --image ghcr.io/acme/aider-bbox:latest --command aider --env OPENAI_API_KEY --mcp
bbox agents doctor aider   # PASS
bbox aider
```
with no manual YAML editing.

HarperZ9:
One small acceptance-criteria suggestion from the agent-observability side: make `agents add/init` emit or preserve a config mutation receipt distinct from `doctor` pass/fail.

Since this command takes user input and writes global config, the receipt would be useful if it captured:

- before/after config fingerprint
- agent name, image, command, egress profile, MCP enabled state, and MCP authz profile
- forwarded env patterns by name only, never values
- validator/schema version used for `ValidateCustomAgent`
- writer behavior: comments preserved, normalized, or intentionally dropped
- explicit non-effects: no workspace `.broodbox.yaml` mutation, no secret values read/written, no credential path widening

[...]

PR #210, opened 2026-07-02
Add bbox agents add/init scaffolding (#201)

JAORMX:
Follow-up to #191 (epic); builds on Phase 1 (#200). Adds the Phase 2 CLI for **authoring** a custom (bring-your-own) agent without hand-editing YAML.

## What's new

- **`bbox agents add NAME --image IMAGE --command CMD [flags]`** — appends a validated custom agent to the **global** config (`~/.config/broodbox/config.yaml` or `--config`), then runs the existing `ValidateCustomAgent` and prints the result. Refuses to overwrite a built-in or an existing custom agent unless `--force`. Flags: `--image`, `--command` (repeatable), `--description`, `--env`, `--env-required`, `--memory`, `--cpus`, `--tmp-size`, `--egress-profile`, `--allow-host`, `--mcp`, `--mcp-authz-profile`, `--force`, `--json`.
- **`bbox agents init [NAME]`** — prints a commented starter `agents:` stanza to stdout, reusing the documentation from `writer.go`'s default template (extracted into a shared `customAgentExampleBlock` so the two can't drift).
- **JSON receipt** on `agents add` and `agents doctor --json` — a re-checkable, paste-safe record: agent summary, env vars **by name only** with present/missing, MCP state, validator version, before/after config SHA-256 fingerprints (add), writer comment-handling note, and explicit non-effects. (Implements @HarperZ9's receipt suggestion.)

## Design notes

- Reuses the pure `config.AgentFromOverride` and `config.ValidateCustomAgent` from #200. New code is CLI (`cmd/bbox`) + a config-file mutation in `internal/infra/config`; the domain stays pure (only a `CustomAgentValidatorVersion` constant added).
- Custom agents remain **global-only** — `agents add` never targets a workspace `.broodbox.yaml`.
- **Comment preservation:** `UpsertAgent` does a comment-preserving YAML node round-trip (encoding the *document* node so leading comments survive). A comment-only file — like the `config init` template — parses to an empty tree, so it is kept verbatim and the new stanza appended; documentation is never dropped. Verified: two sequential `add`s into a fresh `config init` keep all 254 comment lines under a single `agents:` key.

## Acceptance criteria — verified end to end

```bash
bbox agents add aider --image ghcr.io/acme/aider-bbox:latest --command aider --env OPENAI_API_KEY --mcp
bbox agents doctor aider   # [PASS]
```
No manual YAML editing.

## Tests

`task fmt` clean, `task lint` 0 issues, `task test` (race) all 34 packages pass. New tests cover the flag→override mapping, add/doctor end-to-end + JSON receipt, built-in/duplicate refusal, `init` output, and the mutation's comment-preservation / append / force / non-mapping-root cases.


jhrozek:
## Review summary

Ran a full review of this PR (adapted ToolHive review categories to brood-box's DDD layers) using `go-security-reviewer`, `go-architect`, and `go-expert-developer` in parallel, focused on: path-traversal safety for credential/settings paths, secret handling in the new JSON receipt, config-file write safety, egress/MCP privilege defaults, and general Go/test quality.

**Security review**: clean, no findings — `safeRelPath` correctly blocks traversal, the receipt never leaks env values, file perms are correct, egress/MCP defaults are as safe as documented.

**Fixed in follow-up commits**:
- `--mcp-authz-profile` was silently ignored when `--mcp` wasn't also passed — now rejected with a clear error.
- `internal/infra/config/mutate.go`'s `UpsertAgent` did an unlocked read-modify-write, so two concurrent `bbox agents add` calls could silently lose one write. Added a blocking advisory `flock` (matching the existing pattern in `internal/infra/vm/firmware.go`) plus atomic temp-file+rename writes, with a new concurrency regression test.
- `rootMapping` used `yaml.Unmarshal`, which silently drops all but the first document in a `---`-separated multi-document YAML file on re-encode — now detected and rejected.
- `CLAUDE.md`'s custom-agent doc still listed only `list|inspect|doctor`; updated to include `add|init`.

**Deferred, tracked separately**:
- Minor duplication of the "safer defaults" resolution logic (MCP authz / egress profile) across 3-5 call sites in `cmd/bbox` — pure refactor, no functional bug, scoped but not prioritized for this PR.
- **New bug found during that scoping, split out to #211**: a custom agent's own `mcp.enabled: false` override is correctly reflected in `inspect`/`doctor` output but doesn't actually disable the runtime MCP proxy wiring in `main.go`.

All fixes verified via `task fmt && task lint && task test` (race detector on) after each change.

jhrozek (approved):
Reviewed and fixed the issues found (see comment above). LGTM.

PR #212, opened 2026-07-02
Add generic mcp.mode:config declarative MCP injection (#202)

### LINKED ISSUE #202: BYO agents Phase 3: generic `mcp.mode: config` declarative MCP injection
Follow-up to #191 (epic). Phase 1 (#200) added `mcp.mode: env` (the agent discovers the proxy via `BBOX_MCP_URL`). This issue adds the second generic mode, `mcp.mode: config`, so a custom/BYO agent whose harness only reads a config file can be wired to the proxy declaratively — without a per-agent Go `MCPInjector`.

## Scope

Declaratively patch a JSON/JSONC/TOML/YAML config file in the guest home:

```yaml
mcp:
  enabled: true
  mode: config
  inject:
    - guest_path: .config/aider/mcp.json
      format: json
      merge:
        mcpServers:
          broodbox:
            type: streamable-http
            url: "${BBOX_MCP_URL}"
```

- Config shape (`inject` entries: `guest_path`, `format`, `merge` tree) stays **pure domain data** in `pkg/domain/config` / `pkg/domain/agent`.
- The generic merge implementation lives in `internal/infra` — reuse `pkg/clients/internal/configio` (JSON/TOML/YAML merge) and the JSONC support already in `internal/infra/settings`.
- `${BBOX_MCP_URL}` (and other `BBOX_*`) substitution should resolve against the universal env from #200; `config.MCPEndpointPath` remains the single source of truth for the `/mcp` path.
- `ValidateCustomAgent` currently **rejects** `mcp.mode: config` with "not supported in this version" — flip that to accept + validate the `inject` entries (safe relative `guest_path`, known `format`).
- Built-ins keep their bespoke `Plugin.MCPConfig()`; this is only for data-only agents.

## Acceptance criteria

A data-only agent with `mcp.mode: config` boots with its config file patched to point at the in-VM proxy, with no `pkg/clients/` package.

JAORMX:
Closes #202.

BYO agents **Phase 3**. Follow-up to #191 (epic) / #200 (Phase 1 `mcp.mode: env`). Adds the second generic mode, `mcp.mode: config`, so a custom/BYO agent whose harness only reads a config file can be wired to the in-VM MCP proxy declaratively — **without a per-agent Go `MCPInjector`**.

## What it does

Declaratively patch a JSON/JSONC/TOML/YAML config file in the guest home:

```yaml
agents:
  aider:
    image: ghcr.io/example/aider:latest
    command: ["aider"]
    mcp:
      enabled: true
      mode: config
      inject:
        - guest_path: .config/aider/mcp.json
          format: json
          merge:
            mcpServers:
              broodbox:
                type: streamable-http
                url: "${BBOX_MCP_URL}"
```

At VM start the file is created (or merged onto existing content), `${BBOX_*}` references are substituted from the universal env (from #200), and the result is written back in the declared format.

## Layering

- **Config shape stays pure domain data**: `config.MCPInjectConfig` (YAML) → `agent.MCPInjectEntry` (resolved), mapped in `config.AgentFromOverride`. New fields `Agent.MCPInject` and `VMConfig.MCPInject`.
- **Generic merge lives in `internal/infra`**: a new `settings.MCPConfigInjector` reuses the existing `parseConfig`/`serializeConfig`/`deepMerge`/containment helpers used by settings injection (YAML added to that one shared format switch — no duplicate parser). JSONC support is the same `stripJSONC` already in `internal/infra/settings`.
- **`${BBOX_*}` substitution** is scoped to the `BBOX_` prefix so forwarded host secrets can't be expanded into a guest file; unknown/absent tokens are left literal. `config.MCPEndpointPath` (`/mcp`) remains the single source of truth.
- The **VM runner** builds the declarative injector from `VMConfig.MCPInject` when no bespoke plugin injector is set and the proxy is actually available. Built-ins keep their bespoke `Plugin.MCPConfig()`.

## Validation & safety

- `ValidateCustomAgent` now **accepts** `mcp.mode: config` (was rejected as "not supported") and validates each `inject` entry: safe relative `guest_path`, known `format`, non-empty `merge`. `inject` without `mode: config` is rejected; `mode: config` with no `inject` is rejected.
- The load-time "standard profile requires `egress_hosts`" gate is relaxed for `config` mode as it already is for `env` (the proxy is the agent's egress path); `egress.Resolve` remains the authoritative runtime check.
- **Tighten-only**: a workspace-local `.broodbox.yaml` can never introduce or widen `mcp.inject` — it is sourced from the trusted global config only (`mergeAgentOverride`). Locked in by a test.

## Tests

- Domain: inject mapping in `AgentFromOverride`; accept/reject cases in `ValidateCustomAgent`; local-config `mcp.inject` ignored.
- Infra: injector unit tests (JSON create + substitution, merge-into-existing, TOML/YAML, unknown-env-left-literal, path-escape rejected).
- Loader: strict-decode round-trip of the canonical example (free-form `merge` subtree).
- Sandbox: `mcp.mode:config` drops the plugin injector and forwards `MCPInject`, proxy still enabled via `BBOX_MCP_URL`.

`task verify` (fmt + lint + test) is green.


jhrozek:
## Review summary

Reviewed with `go-security-reviewer`, `go-architect`, and `go-expert-developer` in parallel, covering: `${BBOX_*}` substitution safety (can't leak non-BBOX host env/secrets into a guest config file), path containment + symlink safety for the new injector, whether the "standard profile needs egress_hosts" relaxation for `mcp.mode: config` could open unrestricted egress, whether workspace-local `.broodbox.yaml` can introduce/widen `mcp.inject` (it can't — verified against `mergeAgentOverride`), DDD layering, and test coverage.

**Fixed in a follow-up commit**:
- `expandBBOXString` (`internal/infra/settings/mcpinject.go`) paired the first `${` with the *nearest* following `}` via manual `strings.Index` scanning, so an earlier malformed `${...` span in the same string could swallow a legitimate later `${BBOX_MCP_URL}` token — the whole string would round-trip unchanged with no error, silently breaking MCP wiring for any merge-tree value shaped that way. Replaced with a compiled regexp matching well-formed `${IDENTIFIER}` tokens independently, so each token resolves on its own regardless of what precedes it. Added `TestExpandBBOXString`, a direct table-driven test of the function (previously only exercised indirectly via `MCPConfigInjector.Inject`), including the exact regression scenario.

**Everything else checked out clean** — no security findings, no layer violations, injector-selection logic in `runner.go` and helper reuse in `mcpinject.go` are both sound, and the egress relaxation for config mode still hard-restricts to gateway-only.

**Noted but deferred** (same call as on #210): the "is this mode proxy-only (env or config)?" predicate is hand-inlined in 3 places (`custom_agent.go`, two spots in `sandbox.go`) instead of one shared `domainconfig.IsProxyOnlyMode()` helper — pure maintainability, no functional bug, not blocking.

All changes verified via `task fmt && task lint && task test` (race detector on).

jhrozek (approved):
Reviewed and fixed the issue found (see comment above). LGTM.

